PR #44, opened 2020-03-28
update go mod files

xh3b4sd:


pkg/githubclient/types.go
```diff
@@ -53,17 +52,6 @@ func newRepositoryFile(ghContent *github.RepositoryContent) (RepositoryFile, err
...
-	{
-		typ, err := toString(ghContent.Type)
-		if err != nil {
-			return RepositoryFile{}, microerror.Mask(err)
-		}
-
-		if strings.ToLower(typ) != "file" {
-		}
-	}
-
```
kopiczko: What was wrong with that?

xh3b4sd: The linter complained that the one condition was empty. When I removed it `typ` was unused so I had to drop the whole block.

kopiczko: Oh shit yeah. It should return an error. I'm happy we have all that stuff employed.

xh3b4sd: Can you add it back with the error handling in case this is important? I am not sure what kind of error to return. Miss some context here. ❤️

kopiczko: Sure. All good!

PR #48, opened 2020-04-06
Command to automate changes in code for releases

fiunchinho:
This new command tries to automate some of the required steps to [release a new operator version](https://github.com/giantswarm/giantswarm/blob/master/areas/empowerment/versioning-of-operators/how-to-release-an-operator.md). When running from operator folder in the `master` branch, it will:

- Detect what's the current version in `pkg/project/project.go` in the specified folder (default current).
- Change and commit dropping the `-dev` from current version in `pkg/project/project.go`.
- Tag that commit using the version without the `-dev` as tag name.
- Change and commit adding the `-dev` to next patch version in `pkg/project/project.go`.

For example:

- Detect current version is `3.0.4-dev` in `pkg/project/project.go`.
- Change and commit dropping the `-dev` from current version in `pkg/project/project.go`, leaving it like `3.0.4`.
- Tag that commit with the name `3.0.4`.
- Change and commit adding the `-dev` to `pkg/project/project.go` using `3.0.5-dev`.

This is what the commits look like

image

And this is what the new release looks like

image

# Workflow
The workflow when using this command means that the `master` branch contains the work in progress version, something like `1.0.0-dev`. 

[This is somewhat related](https://github.com/giantswarm/giantswarm/issues/9690).

## Releasing a new patch release
When working on a branch which also contains `1.0.0-dev` like master, releasing a new patch version is a matter of running  this command while being on the repository master branch.

```
devctl release
Creating a new release for giantswarm/azure-collector

Current version: 1.0.0-dev
Releasing tag: v1.0.0
Next Work In Progress version: 1.0.1-dev

Github release 'v1.0.0' created successfully! https://github.com/giantswarm/azure-collector/releases/tag/v1.0.0
```

Now, our code would be using `1.0.1-dev` in `pkg/project/project.go`.

## Releasing a new major or minor release
In this case, the branch that we created to work on these changes has already updated the version of the project. If we are using `1.0.0-dev` but changes means a new minor version needs to be released, the branch should reflect the desired version like `1.1.0-dev`.
Same for a new major, like `2.0.0-dev`.

Then, once changes are merged to master, we can just run the command again.

```
devctl release
Creating a new release for giantswarm/azure-collector

Current version: 1.1.0-dev
Releasing tag: v1.1.0
Next Work In Progress version: 1.1.1-dev

Github release 'v1.1.0' created successfully! https://github.com/giantswarm/azure-collector/releases/tag/v1.1.0
```

Now, our code would be using `1.1.1-dev` in `pkg/project/project.go`.


cmd/release/flag.go
```diff
@@ -0,0 +1,119 @@
...
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&f.Organization, "organization", "giantswarm", `Github organization owning the repository.`)
+	cmd.Flags().StringVar(&f.RepositoryName, "repositoryName", "", `Repository name on Github. Defaults to current directory.`)
+	cmd.Flags().StringVar(&f.RepositoryPath, "repositoryPath", "", `Path where the git repository lives in your file system.`)
+}
+
+func (f *flag) Validate() error {
+	var err error
+
+	f.WorkInProgressVersion, f.CurrentVersion, f.NextPatchWorkInProgressVersion, err = getVersions()
```
axbarsan, line 49: Just a nitpick from me:

I'd rather have a single data structure for these versions (a struct), it would be easier to read, and it would be easier to add a new version type in the future

fiunchinho: I see your point but what other new versions would be need? Not sure if grouping them makes sense here.

axbarsan: Maybe needing the previous version for some type of check? (ex: if newer version has a different major). Again, not a blocker from me

cmd/release/runner.go
```diff
@@ -0,0 +1,168 @@
...
+	commit, err := worktree.Commit(commitMessage, &git.CommitOptions{
+		Author: author,
+	})
+
+	return commit, microerror.Mask(err)
+}
+
+func replaceVersionInFile(search, replaceWith string) error {
+	sedCommand := fmt.Sprintf("s/%s/%s/", search, replaceWith)
+	command := exec.Command("sed", "-i", "-E", sedCommand, VersionFile)
```
tuommaki, line 165: I think `-i` here might require some testing / potential workarounds. I have always had problems with it between Mac OS X and rest of the world (mostly GNU sed but not limited to).

See e.g. https://unix.stackexchange.com/a/131940

I'd maybe just implement a function that does that. Just `ReadFile()`, regex replace, `WriteFile()`.

fiunchinho:
> After tagging a release version, it will always bump the patch part of the version?
> 
> So for example we had `1.0.1-dev`, IIUC running the command will:
> 
> * tag `1.0.1`
> * bump to `1.0.2-dev`
> 
> Then if we want to have a new minor/major release, we should bump it manually, at some point, in this case from `1.0.2-dev` to `1.1.0-dev`, before using `devctl release` to release `1.1.0`?

Exactly. How does it sound?

nprokopic:
> > After tagging a release version, it will always bump the patch part of the version?
> > So for example we had `1.0.1-dev`, IIUC running the command will:
> > 
> > * tag `1.0.1`
> > * bump to `1.0.2-dev`
> > 
> > Then if we want to have a new minor/major release, we should bump it manually, at some point, in this case from `1.0.2-dev` to `1.1.0-dev`, before using `devctl release` to release `1.1.0`?
> 
> Exactly. How does it sound?

What about adding a flag something like `--version-bump-option` with possible values `patch` (default), `minor` and `major`?

cmd/release/runner.go
fiunchinho: fixed PTAL

fiunchinho:
> > > After tagging a release version, it will always bump the patch part of the version?
> > > So for example we had `1.0.1-dev`, IIUC running the command will:
> > > 
> > > * tag `1.0.1`
> > > * bump to `1.0.2-dev`
> > > 
> > > Then if we want to have a new minor/major release, we should bump it manually, at some point, in this case from `1.0.2-dev` to `1.1.0-dev`, before using `devctl release` to release `1.1.0`?
> > 
> > 
> > Exactly. How does it sound?
> 
> What about adding a flag something like `--version-bump-option` with possible values `patch` (default), `minor` and `major`?

Actually, that was my first approach. But being able to pass the version to release could lead to situations where we forget to actually change the `pkg/project/project.go` file. Removing the parameter, and letting the command to calculate it forces developers to always use the right version in that file. I think it forces a better workflow.

nprokopic:
> `--version-bump-option` with possible values `patch` (default), `minor` and `major`

Not sure I explained well what I thought. Maybe `--dev-bump-option` is a better name. So if we currently have a version `1.0.2-dev`, then `1.0.2` is released and:
- if you pass `--dev-bump-option patch`, new dev version is `1.0.3-dev`
- if you pass `--dev-bump-option minor`, new dev version is `1.1.0-dev`
- if you pass `--dev-bump-option major`, new dev version is `2.0.0-dev`

sslavic:
How is this automation dealing with updating changelog? There's changelog file in root of most operators + legacy cluster operator had (maybe still has) changelog in source code.

fiunchinho:
> How is this automation dealing with updating changelog? There's changelog file in root of most operators + legacy cluster operator had (maybe still has) changelog in source code.

It's dealing with the changelog, just changing the version in the `pkg/project/project.go` file.

tuommaki:
> How is this automation dealing with updating changelog? There's changelog file in root of most operators...

Starting with recent changes where release components are picked up from `releases` repo, changelogs are dropped from operators altogether. No need to maintain them in operators anymore (unless someone wants to keep CHANGELOG.md up-to-date, but I think that's up to each team I guess).

nprokopic:
> Not sure I explained well what I thought. Maybe `--dev-bump-option` is a better name. So if we currently have a version `1.0.2-dev`, then `1.0.2` is released and:
> 
>     * if you pass `--dev-bump-option patch`, new dev version is `1.0.3-dev`
> 
>     * if you pass `--dev-bump-option minor`, new dev version is `1.1.0-dev`
> 
>     * if you pass `--dev-bump-option major`, new dev version is `2.0.0-dev`

Anyway, not a blocker.

cmd/release/command.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"io"
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/cobra"
+)
+
```
kopiczko, line 11: Problem 1: With that we'll lose ability to review releases before merge.

kopiczko, line 11: Problem 2: How do you do changes to CHAGELOG.md?

fiunchinho: https://github.com/giantswarm/devctl/pull/48/#issuecomment-610309949

fiunchinho: mmm... what's need to be reviewed? I was assuming at this point everything worked fine and the only missing part was creating the tag.

kopiczko: I have mixed feelings about "up to each" team approach. I worry we split too much too soon if we foster that thinking. At least when it comes to the codebases. I'd like to that part to be rather aligned. This is also important when moving from team to team. CC @tuommaki.

kopiczko: E.g:

- I don't want this release yet because I want to merge something.
- I don't want this release because there is a bug.
- We do this release tomorrow because of something.

fiunchinho: Then don't execute this command until

- You merge something missing
- You fixed the bug
- Wait until tomorrow
?

I mean, this command is not automatically executed during a workflow. It's manually executed, each team decides when.

fiunchinho: How can we make it standard for every team?

mig4: Agreed, I missed the news about not maintaining CHANGELOG.md files, and I rather like the practise.

fiunchinho: We need more feedback, good people at @giantswarm/sig-releng

xh3b4sd: The biggest problem I see here is that Jose did something cool for hackathon and now got a downer with all the cautious hesitance. Not saying the good people here caring and holding the line is a bad thing but we should maybe go for it and solve the real problems we have. I do not see a new problem here. All the arguments for not doing it are already true with the manual process. When you still disagree please get together and talk directly in a timeboxed way. Let's make Jose run again. 🚀

xh3b4sd: Extending the changelog with the link in the markdown should be doable in code as well. Would even make it way more reliable because the changelog links are already a little different here and there because humans.

tuommaki: > Agreed, I missed the news about not maintaining CHANGELOG.md files, and I rather like the practise.

After this discussion, I'm super confused. Should they be maintained or not? What tooling uses them and what ensures they are maintained? Yesterday I tried to find some source of truth for the advice of not maintaining versionbundle changelogs, but could only refer the PR comments.

When we need CHANGELOG maintained, having it built-in tooling is important so that it's consistent.

I personally feel that we should automate as much as possible with the release process to avoid problems. And currently we do have problems, because so many things change frequently and it's very difficult to follow and remember every snippet of news post. As I wrote the other day, it's even difficult to find relevant documentation, though this gets improved in a bit.

kopiczko: @tuommaki I think the source of truth for changelogs should be CHANGELOG.md inside the repo. I'm actually quite big on that. To me it doesn't make sense to maintain the changelog outside the repository because it requires extra PR to a separate repo to update it.

One day I talked with @MarcelMue and he said we potentially can link to changelogs inside operator repositories which would be my preference from the development perspective.

I agree there should be more automation around that. And there is work being done towards that direction. There is even `architect/changelog` job. For now it only validates the changelog.

In my ideal world a GitHub release should be created by the CI after new changelog version header is added and the version is somehow injected into the build.

kopiczko: Let's not create drama here.

I will elaborate a bit because I see it isn't obvious.

> hen don't execute this command until
> 
> * You merge something missing
> * You fixed the bug
> * Wait until tomorrow

This is with the assumption you have all that knowledge. If you bypass creating a PR how can you can share you're about to create a release and learn if others don't have anything to say?

I understand where you're coming from. You're a team that pretty much does azure-operator and you can communicate everything internally. But we need to have a way that works for everything else in the company.

This goes pretty much in the direction opposite to https://github.com/giantswarm/giantswarm/issues/9251. With pushing directly to master basically all releases sneak in. No PRs are created so we have zero validation from the CI.

In general pushing directly to master from developer machines is something I'd like to avoid. This sounds completely wrong to me. Ideally that should be blocked on GitHub level. 

Maybe splitting this into two commands that create a PRs for release + create GitHub release and create a PR bumping version is an idea.

Having said that I will disagree and commit that if people like it.

cmd/release/runner.go
```diff
@@ -0,0 +1,231 @@
...
+	search := "## [Unreleased]"
+	replaceWith := fmt.Sprintf("## [Unreleased]\n\n## [%s] %s", r.flag.CurrentVersion, time.Now().Format("2006-01-02"))
+	f, err := ioutil.ReadFile(file)
+	if err != nil {
+		return plumbing.Hash{}, microerror.Mask(err)
+	}
+	filecontents := string(f)
+
+	if !strings.Contains(filecontents, search) {
+		return plumbing.Hash{}, microerror.Maskf(NoUnreleasedWorkFoundInChangelogError, "No unreleased work was found in %s", file)
```
kopiczko, line 199:
```suggestion
		return plumbing.Hash{}, microerror.Maskf(executionFailedError, "no Unreleased header was found in %s", file)
```

cmd/release/runner.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	updatedFileContents := []byte(strings.Replace(filecontents, search, replaceWith, 1))
+	err = ioutil.WriteFile(file, updatedFileContents, 0)
+	if err != nil {
+		return plumbing.Hash{}, microerror.Mask(err)
+	}
+
+	// Change [Unreleased] link
+	m1 := regexp.MustCompile(`(\[Unreleased]:)(.*)(v[0-9]+\.[0-9]+\.[0-9]+)`)
```
kopiczko, line 209:
```suggestion
	m1 := regexp.MustCompile(`^(\[Unreleased]:)(.*)(v[0-9]+\.[0-9]+\.[0-9]+)`)
```

cmd/release/runner.go
```diff
@@ -0,0 +1,231 @@
...
+	updatedFileContents = []byte(m1.ReplaceAllString(string(updatedFileContents), fmt.Sprintf("$1${2}%s$5", r.flag.TagName)))
```
kopiczko, line 210: What is `$5`?

cmd/release/runner.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Change new tag's link
+	taglink := fmt.Sprintf("https://github.com/%s/%s/releases/tag/%s", r.flag.Organization, r.flag.RepositoryName, r.flag.TagName)
+	m2 := regexp.MustCompile(`(\[Unreleased]:.*)`)
+	updatedFileContents = []byte(m2.ReplaceAllString(string(updatedFileContents), fmt.Sprintf("${1}\n[%s]: %s", r.flag.CurrentVersion, taglink)))
+	err = ioutil.WriteFile(file, updatedFileContents, 0)
+	if err != nil {
+		return plumbing.Hash{}, microerror.Mask(err)
+	}
```
kopiczko, line 223: This part we could add in the previous regexp.

cmd/release/runner.go
```diff
@@ -0,0 +1,231 @@
...
+	// Change [Unreleased] link
+	m1 := regexp.MustCompile(`(\[Unreleased]:)(.*)(v[0-9]+\.[0-9]+\.[0-9]+)`)
+	updatedFileContents = []byte(m1.ReplaceAllString(string(updatedFileContents), fmt.Sprintf("$1${2}%s$5", r.flag.TagName)))
+	err = ioutil.WriteFile(file, updatedFileContents, 0)
+	if err != nil {
+		return plumbing.Hash{}, microerror.Mask(err)
+	}
+
+	// Change new tag's link
+	taglink := fmt.Sprintf("https://github.com/%s/%s/releases/tag/%s", r.flag.Organization, r.flag.RepositoryName, r.flag.TagName)
```
kopiczko, line 217: It isn't correct. It should be `compare` URL. See example below. Note first, last and all in the middle.

```
[Unreleased]: https://github.com/giantswarm/architect-orb/compare/v0.8.9...HEAD

[0.8.9]: https://github.com/giantswarm/architect-orb/compare/v0.8.8...v0.8.9
[0.8.8]: https://github.com/giantswarm/architect-orb/compare/v0.8.7...v0.8.8
[0.8.7]: https://github.com/giantswarm/architect-orb/compare/v0.8.6...v0.8.7
[0.8.6]: https://github.com/giantswarm/architect-orb/compare/v0.8.5...v0.8.6
[0.8.5]: https://github.com/giantswarm/architect-orb/compare/v0.8.4...v0.8.5
[0.8.4]: https://github.com/giantswarm/architect-orb/compare/v0.8.3...v0.8.4
[0.8.3]: https://github.com/giantswarm/architect-orb/compare/v0.8.2...v0.8.3
[0.8.2]: https://github.com/giantswarm/architect-orb/compare/v0.8.1...v0.8.2
[0.8.1]: https://github.com/giantswarm/architect-orb/compare/v0.8.0...v0.8.1
[0.8.0]: https://github.com/giantswarm/architect-orb/compare/v0.7.0...v0.8.0
[0.7.0]: https://github.com/giantswarm/architect-orb/compare/v0.6.0...v0.7.0
[0.6.0]: https://github.com/giantswarm/architect-orb/compare/v0.5.3...v0.6.0
[0.5.3]: https://github.com/giantswarm/architect-orb/compare/v0.5.2...v0.5.3
[0.5.2]: https://github.com/giantswarm/architect-orb/compare/v0.5.1...v0.5.2
[0.5.1]: https://github.com/giantswarm/architect-orb/compare/v0.5.0...v0.5.1
[0.5.0]: https://github.com/giantswarm/architect-orb/compare/v0.4.5...v0.5.0
[0.4.5]: https://github.com/giantswarm/architect-orb/compare/v0.4.4...v0.4.5
[0.4.4]: https://github.com/giantswarm/architect-orb/compare/v0.4.3...v0.4.4
[0.4.3]: https://github.com/giantswarm/architect-orb/compare/v0.4.2...v0.4.3
[0.4.2]: https://github.com/giantswarm/architect-orb/compare/v0.4.1...v0.4.2
[0.4.1]: https://github.com/giantswarm/architect-orb/compare/v0.4.0...v0.4.1
[0.4.0]: https://github.com/giantswarm/architect-orb/compare/v0.3.0...v0.4.0
[0.3.0]: https://github.com/giantswarm/architect-orb/compare/v0.2.0...v0.3.0
[0.2.0]: https://github.com/giantswarm/architect-orb/compare/v0.1.2...v0.2.0
[0.1.2]: https://github.com/giantswarm/architect-orb/compare/v0.1.1...v0.1.2
[0.1.1]: https://github.com/giantswarm/architect-orb/compare/v0.1.0...v0.1.1

[0.1.0]: https://github.com/giantswarm/architect-orb/releases/tag/v0.1.0
```

fiunchinho:
- Fixed Changelog modifications.
- Added parameter to enable people to review the release on a branch.

I was thinking if it should clone the remote repository, instead of relying on having the repository cloned locally. What do you think?

xh3b4sd:
> I was thinking if it should clone the remote repository, instead of relying on having the repository cloned locally. What do you think?

I don't understand the difference. Don't you always have to have the repo locally?

cmd/release/runner.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	updatedFileContents := []byte(strings.Replace(filecontents, search, replaceWith, 1))
+	err = ioutil.WriteFile(file, updatedFileContents, 0)
+	if err != nil {
+		return plumbing.Hash{}, microerror.Mask(err)
+	}
+
+	// Update links at the bottom
+	m1 := regexp.MustCompile(`^(\[Unreleased]:)(.*)(v[0-9]+\.[0-9]+\.[0-9]+)(...HEAD)\n`)
```
kopiczko, line 236: There is also a corner case of the first release https://raw.githubusercontent.com/giantswarm/template-operator/master/CHANGELOG.md. But we can iterate on that.

cmd/release/runner.go
```diff
@@ -0,0 +1,249 @@
...
+	err = ioutil.WriteFile(file, updatedFileContents, 0)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (r *runner) addReleaseToChangelog(file string, worktree *git.Worktree) (plumbing.Hash, error) {
+	// Add new entry to changelog
```
kopiczko, line 216: It would be good to have a golden test for changing the changelog.

fiunchinho:
> > I was thinking if it should clone the remote repository, instead of relying on having the repository cloned locally. What do you think?
> 
> I don't understand the difference. Don't you always have to have the repo locally?

I have to have the repo local. But not the command relies on whatever state the repository is in the machine where it's being ran. If we clone the repository on a temporary folder, we get a _clean_ copy.

xh3b4sd:
Got you. When I would run it, I would expect to work with my usual copy of the repo. Nothing hidden. No magic.

cmd/release/runner_test.go
```diff
@@ -0,0 +1,198 @@
+package release
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func Test(t *testing.T) {
```
fiunchinho, line 10: Added a test that checks that changelog manipulation works as expected.

fiunchinho:
We are going with a similar but different approach using Github Workflows/Actions. Thanks everyone.

PR #54, opened 2020-05-11
resource: drop gen resource command

kopiczko:
> TL;TR: I'm killing this. CRUD resource is going to be deprecated.

More details here: https://github.com/giantswarm/giantswarm/issues/5461#issuecomment-626588579


kopiczko:
FYI @giantswarm/sig-operator. Nothing to review here. Only deletions of my own work. I still want to push code generation but I want to invest my time into (hopefully) more fruitful area (I consider this to be CI for now).

PR #46, opened 2020-03-30
gen: my hackathon work to potentially split

kopiczko:

PR #55, opened 2020-05-11
gen: refactor for regeneration and better file overview

kopiczko:
⚠️  Thee is a change to `devctl gen ami` logic. Now package is taken from the `--dir` flag instead of being fixed to `key`.

Here I'm applying lessons learned here https://github.com/giantswarm/devctl/pull/46.

The improvements are not so visible when generating single file but are definitely more visible with multiple. The structure should be also more straightforward when adding new input package.


pkg/gen/input/ami/ami.go
```diff
@@ -0,0 +1,67 @@
...
+		config: config,
+	}
+
+	return f, nil
+}
+
+func (a *AMI) AMIFile() input.Input {
+	a.mustBooted()
+	return file.NewAMIInput(a.params)
+}
```
kopiczko, line 37: I'm not very happy with `mustBooted` but I wanted the interface to be nice. When this is not booted then it's a bug.  It's not the error to handle or print to the end user IMO. Ideally we could do that in `New` but it would be against the principle we follow.

pkg/gen/input/ami/config.go
```diff
@@ -5,9 +5,11 @@ import (
...
 type Config struct {
-	Arch           string
-	Channel        string
-	ChinaDomain    string
+	Arch        string
+	Channel     string
+	ChinaDomain string
+	// Dir is the name of the directory where the files of the resource
+	// should be generated.
 	Dir            string
```
kopiczko, line 13: Added `Dir`. This is standard field for all configs in `input` sub-packages.

pkg/gen/input/ami/funcs.go
```diff
@@ -1,122 +1,67 @@
...
-
-type amiInfo struct {
-	Name string `json:"name"`
-	PV   string `json:"pv"`
-	HVM  string `json:"hvm"`
-}
-
-type amiInfoList struct {
-	AMI []amiInfo `json:"amis"`
-}
```
kopiczko, line 55: This got moved to the `scrapeVersionAMI` body. It's only used there.

pkg/gen/input/ami/funcs.go
```diff
@@ -1,122 +1,67 @@
 package ami
```
kopiczko, line 1: Only the scraping part is left in this file. It gets the values to populate template data.

pkg/gen/input/ami/internal/file/ami.go
```diff
@@ -0,0 +1,35 @@
+package file
```
kopiczko, line 1: This is a file template file. It's clean and simple. Shows the template and the parameters it uses.

pkg/gen/input/ami/funcs.go
```diff
@@ -175,6 +115,16 @@ func scrapeVersionAMI(source io.Reader) (map[string]string, error) {
...
 
+	type amiInfo struct {
+		Name string `json:"name"`
+		PV   string `json:"pv"`
+		HVM  string `json:"hvm"`
+	}
+
+	type amiInfoList struct {
+		AMI []amiInfo `json:"amis"`
+	}
```
kopiczko, line 126: Moved here.

pkg/gen/input/ami/funcs.go
```diff
@@ -188,19 +138,3 @@ func scrapeVersionAMI(source io.Reader) (map[string]string, error) {
...
-var amiJSON = []byte({{ .AMIInfoString }})
-var amiInfo = map[string]map[string]string{}
-
-func init() {
-	err := json.Unmarshal(amiJSON, &amiInfo)
-	if err != nil {
-		panic(err)
-	}
-}
-`
```
kopiczko, line 206: This is now in `file` package to decouple templating from computing template data.

xh3b4sd:
Build fails.

pkg/gen/input/ami/internal/file/ami.go
```diff
@@ -0,0 +1,35 @@
...
+	return i
+}
+
+var amiTemplate = `// NOTE: This file is generated by devctl. Do not edit.
+package {{ .Package }}
+
+import "encoding/json"
+
+var amiJSON = []byte({{ .AMIInfoString }})
+var amiInfo = map[string]map[string]string{}
```
xh3b4sd, line 27: I saw the rendered file and thought this reads way better when having to debug and understand things in the repo. 

```suggestion
var amiInfo = map[string]map[string]string{}

var amiJSON = []byte({{ .AMIInfoString }})
```

PR #56, opened 2020-05-28
Generate Makefiles

njuettner:
Add to generate Makefiles for Golang projects.

This is a followup from the conversation with Pawel about adding a Makefile for the aws-operator (https://github.com/giantswarm/aws-operator/pull/2445). I hope this PR is a better approach for using Makefiles 😄.

But again this PR is open for discussions.


cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
...
+type TemplateConfig struct {
+	Application string
+}
+
+var Makefile = `APPLICATION   ?= {{.Application}}
+VERSION       ?= $(shell git describe --tags --always --dirty)
+SOURCES       = $(shell find . -name '*.go')
+GOPKGS        = $(shell go list ./...)
+TAG           ?= $(VERSION)
+BUILD_FLAGS   ?= -v
```
kopiczko, line 13: I'd skip the `-v` flag. It's noisy.

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
+package template
+
+//TemplateConfig inherits all placeholder which should be overwritten.
+type TemplateConfig struct {
+	Application string
+}
+
+var Makefile = `APPLICATION   ?= {{.Application}}
```
kopiczko, line 8:
```suggestion
var Makefile = `# DO NOT EDIT. Generated with:
#
#    devctl gen makefile
#

APPLICATION   ?= {{.Application}}
```

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
+package template
+
+//TemplateConfig inherits all placeholder which should be overwritten.
+type TemplateConfig struct {
+	Application string
+}
+
```
kopiczko, line 7: Because of consistency I would be happier if we followed the pattern like [here](https://github.com/giantswarm/devctl/blob/master/pkg/gen/input/ami/internal/file/ami.go).

Even though I understand it may feel like an overkill for this particular case.

If you'd like to do so then we need to make this file regenerable [here](https://github.com/giantswarm/devctl/blob/c5630628e98c5fb33582f1511a8f02332c90b50d/pkg/gen/gen.go#L44-L52).

But I won't block on that because I think it's already an improvement.

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
...
+var Makefile = `APPLICATION   ?= {{.Application}}
+VERSION       ?= $(shell git describe --tags --always --dirty)
```
kopiczko, line 9:
```suggestion
VERSION       ?= $(shell architect project version)
```

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
...
+
+.PHONY: lint
+## lint: runs golangci-lint
+lint:
+	golangci-lint run --timeout=15m ./...
+
+.PHONY: test
+## test: runs go test with default values
+test:
+	go test -v -race -cover $(GOPKGS)
```
kopiczko, line 55: I'd drop `-v` here as well. We've been in a place where we had so many packages that it was difficult to find what was failing when tests were red.

cmd/gen/makefile/flag.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type flag struct {
+	App    string
+	Output string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&f.App, app, "", `Name of the application, this will be the name of the go binary.`)
+	cmd.Flags().StringVar(&f.Output, "output", "", `Where you want the generated Makefile to be saved`)
```
kopiczko, line 20: That flat I'd skip unless there is a use-case for that. We should generate it as `Makefile` in the current dir.

cmd/gen/makefile/template/template.go
kopiczko, line 13: It would be good to make it on par with the [orb](https://github.com/giantswarm/architect-orb/blob/7e3e1243d49a67d6f07482e40145e3da3598ded2/src/commands/go-test.yaml#L5-L9). But it's fine with me if we don't do this in first iteration.

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
+package template
+
+//TemplateConfig inherits all placeholder which should be overwritten.
+type TemplateConfig struct {
+	Application string
+}
+
+var Makefile = `APPLICATION   ?= {{.Application}}
+VERSION       ?= $(shell git describe --tags --always --dirty)
+SOURCES       = $(shell find . -name '*.go')
```
kopiczko, line 10: Do we need sources when we have only phony targets?

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
...
+GOPKGS        = $(shell go list ./...)
+TAG           ?= $(VERSION)
```
fiunchinho, line 12: Why have both `VERSION` and `TAG`?

cmd/gen/makefile/template/template.go
```diff
@@ -0,0 +1,67 @@
+package template
```
fiunchinho, line 1: What do you folks think about running `go mod tidy` automagically after building?

fiunchinho:
This looks great! It would be cool to try to replicate [our orb](https://github.com/giantswarm/architect-orb/blob/a4a74fed9a3a719ed7e62b0c16bcb866a08838d1/src/commands/go-test.yaml) as much as possible, so that we can later use the Makefile from the orb instead.

cmd/gen/makefile/runner.go
```diff
@@ -0,0 +1,64 @@
...
+	err = makefileInput.Params(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	_, err = os.Create("Makefile")
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = gen.Execute(
```
njuettner, line 50: @kopiczko I moved the logic to your suggested place. How can I avoid getting the `zz_generated.Makefile` file and get a `Makefile` file? TBH I underestimated the complexity of this logic and I'm not sure how to make this right 🙂 . I would appreciate your input.

kopiczko: How about that https://github.com/giantswarm/devctl/pull/57 ?

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return i
+}
+
+var makefileTemplate = `# DO NOT EDIT. Generated with:
+#
+#    devctl gen makefile
+#
+
+APPLICATION    ?= {{.Application}}
```
kopiczko, line 25:
```suggestion
APPLICATION    ?= $(shell basename $(go list .))
```

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,84 @@
...
+VERSION        ?= $(shell architect project version)
+GITSHA1        = $(shell git rev-parse --verify HEAD)
+BUILDTIMESTAMP = $(shell date -u '+%FT%TZ')
+LDFLAGS        ?= -w -linkmode 'auto' -extldflags '-static' \
+  -X '$(go list .)/pkg/project.buildTimestamp=${BUILDTIMESTAMP}' \
+  -X '$(go list .)/pkg/project.gitSHA=${GITSHA1}'
```
kopiczko, line 31:
```suggestion
  -X '$(go list .)/pkg/project.gitSHA=${GITSHA1}' \
  -X '$(go list .)/pkg/project.version=${VERSION}'
```

njuettner:
I cleaned it up a bit and dropped the flag since we don't need it anymore.

Had to change `$(shell basename $(go list .))`to  `$(shell basename $(shell go list .))` other wise it was complaining on my Mac that basename not getting the string.

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,82 @@
...
+
+.PHONY: build-linux
+## build-linux: builds binary for linux/amd64
+build-linux: clean
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build .
+
+.PHONY: build-darwin
+## build-darwin: builds binary for darwin/amd64
+build-darwin: clean
+	GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 go build .
```
kopiczko, line 45:
```suggestion
.DEFAULT_GOAL := build

.PHONY: build build-darwin build-linux
## build: builds a local binary
build: $(APPLICATION)
## build-darwin: builds a local binary for darwin/amd64
build-darwin: $(APPLICATION)-darwin
## build-linux: builds a local binary for linux/amd64
build-linux: $(APPLICATION)-linux

$(APPLICATION): $(APPLICATION)-$(OS)
	cp -a $< $@

$(APPLICATION)-%: $(SOURCES)
	GOOS=$* GOARCH=amd64 go build -ldflags "$(LDFLAGS)" -o $@ .
```

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,82 @@
...
+
+.PHONY: install
+## install: install the application
+install:
+	go install .
```
kopiczko, line 50:
```suggestion
	go install -ldflags "$(LDFLAGS)" .
```

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,82 @@
...
+
+.PHONY: run
+## run: runs go run main.go
+run:
+	go run -race .
```
kopiczko, line 55:
```suggestion
	go run -ldflags "$(LDFLAGS)" -race .
```

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,82 @@
...
+
+.PHONY: clean
+## clean: cleans the binary
+clean:
+	go clean
```
kopiczko, line 60:
```suggestion
	rm -f $(APPLICATION)*
	go clean
```

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,82 @@
...
+
+.PHONY: lint
+## lint: runs golangci-lint
+lint:
+	golangci-lint run -E gosec -E goconst --timeout=15m ./...
+
+.PHONY: test
+## test: runs go test with default values
+test:
+	go test -ldflags $(LDFLAGS) -race -cover ./...
```
kopiczko, line 70:
```suggestion
	go test -ldflags "$(LDFLAGS)" -race -cover ./...
```

pkg/gen/input/makefile/internal/file/makefile.go
njuettner: @kopiczko I tried that on my mac, `make build-linux` and `make build-darwin` just works but it doesn't work with `make build` or `make` directly.

```
cmd/go: unsupported GOOS/GOARCH pair .o/amd64
make: *** [devctl-.o] Error 2
```

kopiczko: Can you paste the full outpu?

pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -0,0 +1,84 @@
...
+
+.PHONY: lint
+## lint: runs golangci-lint
+lint:
+	golangci-lint run -E gosec -E goconst --timeout=15m ./...
+
+.PHONY: test
+## test: runs go test with default values
+test:
+	go test -ldflags "$(LDFLAGS)" -race -cover ./...
```
kopiczko, line 72: Do we need `-cover` flag there? We don't use it in the CI and I'm not sure it's useful as we are not super fussed about test coverage in general. We also don't have `-race` but I think that's ok if it doesn't take ages to run it in the CI machine.

https://github.com/giantswarm/architect-orb/blob/d1535b34fad4df844df9f183325fde5c24b28ba8/src/commands/go-test.yaml#L23

njuettner: @kopiczko I dropped it but left -race. I think this shouldn't increase the amount of time when running it.

kopiczko: Fine with me.

PR #61, opened 2020-06-15
workflows: add command

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/8335

Here's how final product would look like: https://github.com/kopiczko/test-gh-workflows/tree/5ef09186c61133ec3c5250423879f7887207baa4/.github/workflows

### DIFF
--- a/cmd/gen/workflows/flag.go
+++ b/cmd/gen/workflows/flag.go
@@ -1,7 +1,6 @@
 package workflows
 
 import (
-	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/gen/input/workflows/internal/file/create_release.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+func NewCreateReleaseInput(p params.Params) input.Input {
+	i := input.Input{
+		Path:         params.RegenerableFileName(p, "create_release.yaml"),
+		TemplateBody: createReleaseTemplate,
+		TemplateDelims: input.InputTemplateDelims{
+			Left:  "##-ignore-left-##",
+			Right: "##-ignore-right-##",
+		},
```
kopiczko, line 15: I needed to add this because there is a lot of `${{ ... }}` in the templates and I don't want to change that because we don't really use templating here.

cmd/gen/workflows/flag.go
```diff
@@ -0,0 +1,26 @@
...
+const (
+	flagDir = "dir"
+)
+
+type flag struct {
+	Dir string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&f.Dir, flagDir, "./.github/workflows", `Directory where the generated code should be located.`)
```
njuettner, line 17: I wonder if this will change? or if we can drop it similar to Makefile

kopiczko: I dropped this flag.

kopiczko:
Looks like [quay has issues again](https://gigantic.slack.com/archives/CDJJ17U81/p1592484384113100). I tested it locally. Going with it.

PR #64, opened 2020-06-23
workflows: support minor long-lived release branches

kopiczko:
"Create Release PR" workflow will now be triggered for branches like:

```
release-v1.2.x#release#v1.2.3
```

But wont be triggered for branches like:

```
release-v1.x.x#release#v1.2.3
```

### DIFF
--- a/pkg/gen/input/workflows/internal/file/create_release_pr.go
+++ b/pkg/gen/input/workflows/internal/file/create_release_pr.go
@@ -28,7 +28,7 @@ on:
     branches:
       - 'master#release#v*.*.*'
       - 'legacy#release#v*.*.*'
-      - 'release-v*.x.x#release#v*.*.*'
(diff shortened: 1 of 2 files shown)


pkg/gen/input/workflows/internal/file/create_release_pr.go
```diff
@@ -28,7 +28,7 @@ on:
     branches:
       - 'master#release#v*.*.*'
       - 'legacy#release#v*.*.*'
-      - 'release-v*.x.x#release#v*.*.*'
+      - 'release-v*.*.x#release#v*.*.*'
```
kopiczko, line 31: To be even more rigid:

```suggestion
      - 'release-v*.*.x#release#v*.*.*'
      # "!" negates previous positive patterns so it has to be at the end.
      - '!release-v*.x.x#release#v*.*.*'
```

PR #67, opened 2020-06-25
gen/command: add

kopiczko:


kopiczko:
Won't pursue that for now.

PR #68, opened 2020-06-25
CLI for release management

tfussell:
Implementation of https://github.com/giantswarm/giantswarm/issues/11782


tfussell:
This isn't perfect, but it should still save some time checking release versions and wiring everything up. We would like to use this for creating temporary releases directly on CPs so work will continue as part of that (including tests hopefully).

cmd/create/release/flag.go
```diff
@@ -0,0 +1,55 @@
...
+type flag struct {
+	Base       string
+	Components []string
+	Name       string
+	Overwrite  bool
+	Provider   string
+	Releases   string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
```
tfussell, line 28: Flags are defined here.

pkg/changelog/changelog.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Data about a particular component version returned from parsing a changelog
+type Version struct {
+	Link    string
+	Name    string
+	Content string
+}
+
+func ParseChangelog(componentName, componentVersion string) (*Version, error) {
```
tfussell, line 90: This is one of the core functions. Returns the changes for a specific version of an upstream component by parsing its changelogs.

pkg/release/archive.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/giantswarm/microerror"
+)
+
+func ArchiveRelease(name, releases, provider string) error {
```
tfussell, line 11: This is an extra command I added to move an old release into `archived/` and deregister it.

pkg/release/create.go
```diff
@@ -0,0 +1,99 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
+	"github.com/giantswarm/microerror"
+	"sigs.k8s.io/yaml"
+)
+
+func CreateRelease(name, base, releases, provider string, components []string, overwrite bool) error {
```
tfussell, line 15: The main entry point for creating a release is here.

pkg/changelog/changelog.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+// Regex patterns used by all Giant Swarm components
+const (
+	commonStartPattern = "(?m)^## \\[v?(?P<Version>\\d+\\.\\d+\\.\\d+)\\].*(?P<Date>\\d{4}-\\d{2}\\-\\d{2})$"
+	commonEndPattern   = "(?m)^\\[.*\\]:.*$"
+)
+
+// Parameters defining how to parse and extract release info about all known components
+var knownComponentParseParams = map[string]parseParams{
```
MarcelMue, line 23: Fine for now but this should live somewhere easily findable.

pkg/changelog/changelog.go
MarcelMue: Dios mio - this is certainly a function. Fine for now but maybe we can abstract some parts of it.

cmd/create/command.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/devctl/cmd/create/release"
+)
+
+const (
+	name        = "create"
+	description = "Create resources in the current working directory."
```
MarcelMue, line 16: Some examples would be neato here because the combination of flags can be quite complex from what I see.

cmd/archive/release/flag.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/Masterminds/semver/v3"
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	nameFlag     = "name"
+	providerFlag = "provider"
+	releasesFlag = "releases"
```
kopiczko, line 13: Let's stick to convetion:

```suggestion
	flagName     = "name"
	flagProvider = "provider"
	flagReleases = "releases"
```

cmd/create/release/flag.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	baseFlag      = "base"
+	componentFlag = "component"
+	nameFlag      = "name"
+	overwriteFlag = "overwrite"
+	providerFlag  = "provider"
+	releasesFlag  = "releases"
```
kopiczko, line 16: Same here.

pkg/changelog/changelog.go
```diff
@@ -0,0 +1,254 @@
+package changelog
```
kopiczko, line 1: Can we put that under `pkg/release` to indicate this is releases' changlog?

pkg/release/create.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Define release CR
+	var updatesRelease v1alpha1.Release
+	newVersion := *semver.MustParse(name) // already validated to be a valid semver string
+	updatesRelease.Name = "v" + newVersion.String()
+	for _, componentVersion := range components {
+		split := strings.Split(componentVersion, "@")
+		updatesRelease.Spec.Components = append(updatesRelease.Spec.Components, v1alpha1.ReleaseSpecComponent{
+			Name:    split[0],
+			Version: split[1],
```
yulianedyalkova, line 32: I didn't see any validation that the components have the `component@version` format so this will  panic with `slice bounds out of range` if someone forgets to provide the `@version`.

MarcelMue:
> I have only some structure level comments here. I don't dive deeply into implantation details as I don't deal with releases often enough.
> 
> I'd also suggest to change the command structure to:
> 
> ```
> devctl release archive
> devctl release create
> ```
> 
> The reason for that is that it's easier to discover functionality with shell completion that way. `devctl` is a swiss army knife tool and IMO some common verbs are not good enough entry points when we don't simply deal with REST API.

I disagree with the structure suggestion here. Discoverability is always an issue but the structure of `cmd verb object` is used in basically all CLIs (`kubectl get nodes`, `gsctl list clusters`, ...) changing this here seems questionable to me.

kopiczko:
> > I have only some structure level comments here. I don't dive deeply into implantation details as I don't deal with releases often enough.
> > I'd also suggest to change the command structure to:
> > ```
> > devctl release archive
> > devctl release create
> > ```
> > 
> > 
> > The reason for that is that it's easier to discover functionality with shell completion that way. `devctl` is a swiss army knife tool and IMO some common verbs are not good enough entry points when we don't simply deal with REST API.
> 
> I disagree with the structure suggestion here. Discoverability is always an issue but the structure of `cmd verb object` is used in basically all CLIs (`kubectl get nodes`, `gsctl list clusters`, ...) changing this here seems questionable to me.

- We already had this discussion when `devctl repo list` was implemented.
- ` docker image ls`
- `kubectl` operates mostly on dynamic REST API so this structure makes more sense here.

xh3b4sd:
Hi gang. Can we get over with this? I see people crafting releases and thought some diffing would be cool to have for new releases, especially because we have that in releases@master already. Let's iterate if there is more to it and let me know when I can do anything to get this going.

tfussell:
> Hi gang. Can we get over with this? I see people crafting releases and thought some diffing would be cool to have for new releases, especially because we have that in releases@master already. Let's iterate if there is more to it and let me know when I can do anything to get this going.

I've been focused on KVM 12.0.0, planning to finish this up today.

pkg/changelog/changelog.go
tfussell: Indeed. I cleaned it up a bit. There's plenty of room for improvement (and unit tests), but it does work for the most part.

pkg/changelog/changelog.go
tfussell: Agreed. We can probably have a few generic parsers that handle different styles of changelog. Keep a Changelog-style, Kubernetes-style, Container Linux-style, etc.

cmd/create/command.go
tfussell: I added an example.

cmd/archive/release/flag.go
tfussell: Didn't know, thanks.

pkg/changelog/changelog.go
tfussell: Done.

pkg/release/create.go
tfussell: I added some validation that apps have 2-3 parts separated by "@" symbols and components have 2 or an error is returned. Thanks for the extra push to make things nicer for users.

tfussell:
> I have only some structure level comments here. I don't dive deeply into implantation details as I don't deal with releases often enough.
> 
> I'd also suggest to change the command structure to:
> 
> ```
> devctl release archive
> devctl release create
> ```
> 
> The reason for that is that it's easier to discover functionality with shell completion that way. `devctl` is a swiss army knife tool and IMO some common verbs are not good enough entry points when we don't simply deal with REST API.

If it's alright with you, I would like to stick with the current approach (`<verb> <noun>`) since that's what I've already written and it's more consistent with our other command. Afterwards, I would then try to come up with a documented standard for all of the GS CLIs (or at least a heuristic for when to choose one style or the other) and add it to `fmt`. What do you think?

kopiczko:
This will be inconsistent with `devctl repo list` already. We now also have `opsctl release deploy`. I can only disagree and commit here.

tfussell:
> This will be inconsistent with `devctl repo list` already. We now also have `opsctl release deploy`. I can only disagree and commit here.

I discussed a bit with SIG-UX. The conclusion was basically that we should prefer `<verb> <noun>` for consistency and that `<noun> <verb>` is okay when it makes more sense. I've adjusted the order now.

PR #63, opened 2020-06-22
makefile: generate packaging scripts for CLIs

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/11715

This adds a new `-f|--flavour` flag to the `devctl gen makefile` command, with the supported values being `app|cli|operator`.

Using the new flag, we are able to generate different Makefile scripts for different project types. For example, this PR adds scripts for packaging, for the CLI flavour of the Makefile.

The generated files are `.tar.gz` archives, for each supported platform (MacOS and generic Linux), which contain the executable binary, README and LICENSE. They are built in a Docker container.

PR #69, opened 2020-06-29
Add 'v' prefix to binary name

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/11783

This PR:
* Adds a 'v' prefix to all version numbers in the generated Makefile/GitHub workflows. (e. g. `some-operator-1.0.0-darwin.tar.gz` now becomes `some-operator-v1.0.0-darwin.tar.gz`
* Refactors the generated Makefile, so it is more extendable, as suggested [here](https://github.com/giantswarm/devctl/pull/63#discussion_r445524976).

### DIFF
--- a/pkg/gen/input/workflows/internal/file/create_release.go
+++ b/pkg/gen/input/workflows/internal/file/create_release.go
@@ -258,7 +258,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GO_VERSION: 1.14.2
       ARTIFACT_DIR: bin-dist
-      PKG_VERSION: v${{ needs.gather_facts.outputs.version }}
+      TAG: v${{ needs.gather_facts.outputs.version }}
     needs:
       - create_release
       - gather_facts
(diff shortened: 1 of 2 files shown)


pkg/gen/input/workflows/internal/file/create_release.go
```diff
@@ -258,7 +258,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GO_VERSION: 1.14.2
       ARTIFACT_DIR: bin-dist
-      PKG_VERSION: ${{ needs.gather_facts.outputs.version }}
+      PKG_VERSION: v${{ needs.gather_facts.outputs.version }}
```
kopiczko, line 261: I like to think that the version is the thing w/o "v". And it's usually easier to add prefix than remove it in scripts so I think it would be good to keep the naming aligned.

```suggestion
      TAG: v${{ needs.gather_facts.outputs.version }}
```

pkg/gen/input/workflows/internal/file/create_release.go
```diff
@@ -290,7 +290,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
         with:
-          ref: v${{ needs.gather_facts.outputs.version }}
+          ref: ${{ env.PKG_VERSION }}
```
kopiczko, line 293:
```suggestion
          ref: ${{ env.TAG }}
```

PR #70, opened 2020-07-01
makefile: remove all the version metadata from the binary name

axbarsan:
This is the last one, I promise 🤞 

With this, built binaries in the archived packages will have a simpler name:
```diff
- kubectl-gs-v0.5.0-darwin-amd64
+ kubectl-gs
```


pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -75,7 +75,7 @@ package-linux: $(PACKAGE_DIR)/$(APPLICATION)-v$(VERSION)-linux-amd64.tar.gz
 $(PACKAGE_DIR)/$(APPLICATION)-v$(VERSION)-%-amd64.tar.gz: DIR=$(PACKAGE_DIR)/$<
 $(PACKAGE_DIR)/$(APPLICATION)-v$(VERSION)-%-amd64.tar.gz: $(APPLICATION)-v$(VERSION)-%-amd64
 	mkdir -p $(DIR)
-	cp $< $(DIR)
+	cp $< $(DIR)/$(APPLICATION)
```
kopiczko, line 78: Do we need `$(APPLICATION)-v$(VERSION)-%-amd64` target at all then? I feels like this one can depend on `$(APPLICATION)-%` now

PR #71, opened 2020-07-02
Add workflow for creating release branch when new minor is tagged

stone-z:
Toward https://github.com/giantswarm/giantswarm/issues/10828

This workflow creates a new release branch when:
- a new version (`v*.*.*`) is tagged
- the tag introduces a new major or minor version ( > major/minor version than previous tag)
- the branch does not already exist

This breaks when the previous tag is not valid semver but I think that's ok.

The workflow includes a python script. This is because:
- GitHub string sanitization breaks doing things like `HEAD^` when using the git cli directly
- the equivalent bash to do this is super ugly
- maintaining a separate build tool for this seemed like overkill
- this is (hopefully) unlikely to change often

I've tested this in a personal repo. Hoping to get feedback up front, but won't merge until I try this on one of our repos for a little bit.

Sample output:
Screenshot from 2020-07-02 18-12-30
Screenshot from 2020-07-02 18-11-30


pkg/gen/input/workflows/internal/file/create_release_branch.go
```diff
@@ -0,0 +1,136 @@
...
+        previous_branch_name = "release-v{}.{}.x".format(parent_version.major, parent_version.minor)
+
+        # Check if the current tag version introduces a new major or minor version
+        new_version = False
+        if current_version.major > parent_version.major:
+          print("Current tag is a new major version")
+          new_version = True
+        elif current_version.major == parent_version.major and current_version.minor > parent_version.minor:
+          print("Current tag is a new minor version")
+          new_version = True
```
kopiczko, line 102: I'm not sure what's going on here but before I dive in I'd like to clear something.

For a given tag `v1.2.3` we want to make sure that branch `release-v1.2.x` exists correct?

If so why we can't just take the tag, transform its name into the desired branch name and create the branch if it doesn't exist?

stone-z: We agreed to create the branch only once a _new_ minor (or major) is tagged. So if we had `v1.2.3` already, the branch `release-1.2.x` would be created when `v1.3.0` is tagged (or `v2.0.0`). So since this action runs on the "new" tag, it has to look backwards at the "old" version to know what to name the branch

pkg/gen/input/workflows/internal/file/create_release_branch.go
```diff
@@ -0,0 +1,136 @@
...
+        print("Release branch does not exist")
+
+        # Create the branch
+        print("Creating release branch {}".format(previous_branch_name))
+        origin = repo.remote()
+
+        # Create a local branch
+        release_branch = repo.create_head(previous_branch_name)
+        # Check it out
+        release_branch.checkout()
```
stone-z, line 127: I noticed I'm currently creating the branch from the tagged commit, not the parent. Will fix this

pkg/gen/input/workflows/internal/file/create_release_branch.go
```diff
@@ -0,0 +1,136 @@
...
+
+# Creates a branch for the 'previous' minor version when a 'new' minor version is tagged
+
+name: create-minor-version-branch
+
+on:
+  push:
+    tags: ['v*.*.*']
+
+jobs:
```
kopiczko, line 30: I'd like to add that to every workflow we have here because it makes further debugging and improvements easier.

Could you also show me the output of that job for and example workflow run? I have a hunch we can simplify the code from `create-release-branch` a lot.

```suggestion
jobs:
  debug_info:
    name: Debug info
    runs-on: ubuntu-18.04
    steps:
      - name: Print github context JSON
        run: |
          cat <<EOF
          ${{ toJson(github) }}
          EOF
```

pkg/gen/input/workflows/internal/file/create_release_branch.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"github.com/giantswarm/devctl/pkg/gen/input"
+	"github.com/giantswarm/devctl/pkg/gen/input/workflows/internal/params"
+)
+
+func NewCreateReleaseBranchInput(p params.Params) input.Input {
+	i := input.Input{
+		Path:         params.RegenerableFileName(p, "create_release_branch.yaml"),
+		TemplateBody: createReleaseBranchTemplate,
```
kopiczko, line 11:
```suggestion
		TemplateBody: createReleaseBranchTemplate,
		TemplateDelims: input.InputTemplateDelims{
			Left:  "{{{{",
			Right: "}}}}",
		},
```

pkg/gen/input/workflows/internal/file/create_release_branch.go
```diff
@@ -0,0 +1,136 @@
...
+#    devctl gen workflows
+#
+
+# Creates a branch for the 'previous' minor version when a 'new' minor version is tagged
+
+name: create-minor-version-branch
+
+on:
+  push:
+    tags: ['v*.*.*']
```
kopiczko, line 28: One more thing to check is if this workflow will be run at all if the tag is created by another workflow. Apparently GH workflows are not triggered by events from other workflows. See https://github.com/giantswarm/kubectl-gs/pull/64

pkg/gen/input/workflows/internal/file/create_release_branch.go
stone-z: Sure, what are you looking for? Here's the likely relevant part, but if you're hoping GH provides the previous tag or something, I didn't find one.

```
{
  "token": "***",
  "job": "debug_info",
  "ref": "refs/tags/v3.5.0",
  "sha": "397fb75698f44ea5f3f289f17e860204da89e8c3",
  "repository": "stone-z/create-minor-branch-action",
  "repository_owner": "stone-z",
  "repositoryUrl": "git://github.com/stone-z/create-minor-branch-action.git",
  "run_id": "156462941",
  "run_number": "39",
  "actor": "stone-z",
  "workflow": "create-minor-version-branch",
  "head_ref": "",
  "base_ref": "",
  "event_name": "push",
  "event": {
    "after": "397fb75698f44ea5f3f289f17e860204da89e8c3",
    "base_ref": "refs/heads/master",
    "before": "0000000000000000000000000000000000000000",
    "commits": [],
    "compare": "https://github.com/stone-z/create-minor-branch-action/compare/v3.5.0",
    "created": true,
    "deleted": false,
    "forced": false,
    "head_commit": {
      "author": {
        <redacted, but it's me>
      },
      "committer": {
        <redacted, but it's me>
      },
      "distinct": true,
      "id": "397fb75698f44ea5f3f289f17e860204da89e8c3",
      "message": "Original with refactor",
      "timestamp": "2020-07-03T12:26:54+02:00",
      "tree_id": "eca42b58132a65e7f095ab530d0521a9760b8e9e",
      "url": "https://github.com/stone-z/create-minor-branch-action/commit/397fb75698f44ea5f3f289f17e860204da89e8c3"
    },
    "ref": "refs/tags/v3.5.0",
    "repository": {
      "created_at": 1592916067,
      "default_branch": "master",
	... a bunch of junk ...
```

pkg/gen/input/workflows/internal/file/create_release_branch.go
kopiczko: I'm pretty sure it won't be triggered when tag is created by another workflow https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

So you'd need to glue that job to `Create Release` workflow I guess.

stone-z: Yeah you're right, so to have any separate workflows after `Create Release` we'd have to make it use an OAuth token instead. Do you have a problem gluing this to Create Release?

stone-z:
Similar to the 'workflows triggering workflows" issue, a workflow using the `GITHUB_TOKEN` can't modify a workflow file unless it is exactly the same as a workflow already on a different branch. [Related issue](https://github.community/t/refusing-to-allow-a-bot-to-create-or-update-workflow-when-pushing-tags-on-older-commits/17786)

This means if we have `v1` and `v2` comes along having modified the workflow, this action can't create the branch from the parent commit without using a PAT.

Any ideas for alternatives? If not, I'll have to create an OAuth token for this purpose.

pkg/gen/input/workflows/internal/file/create_release_branch.go
kopiczko: > Do you have a problem gluing this to Create Release?

Not at all.

kopiczko:
> a workflow using the GITHUB_TOKEN can't modify a workflow file

Why would we want to do that?

stone-z:
> > a workflow using the GITHUB_TOKEN can't modify a workflow file
> 
> Why would we want to do that?

Ok, I have since tried a bunch more stuff and this should not be a blocker after all.

We aren't actually modifying the workflow, but for whatever reason GitHub sees it that way IF two sequential commits _both_ modify the workflow and then get tagged. I can not make any sense of why this is, but in practice I think it'll be fairly uncommon that we modify and immediately tag a workflow file twice in a row.

So we should be good to go without the OAuth token. Will update this PR with the requested changes momentarily

(in case someone wants it)
Full repro scenario:
1. Commit `abc1` made some change to the workflow
2. `abc1` was tagged `v1` and this automation pushed a branch `release-v1`
3. The NEXT commit `abc2` made some change to the workflow
4. `abc2` is tagged `v2` and the action tries to run --> fails with the message:
` ! [remote rejected] release-v2 -> release-v2 (refusing to allow a GitHub App to create or update workflow '.github/workflows/create-minor-release-branch.yml' without 'workflows' permission)`

If there is an additional non-tagged commit in between, GitHub does not complain

pkg/gen/input/workflows/internal/file/create_release_branch.go
stone-z: Added

pkg/gen/input/workflows/internal/file/create_release_branch.go
stone-z: Added the debug info job to the workflow, but I guess this will already be present in `Create Release` anyway

kopiczko:
Sorry but I'm lost. I think I don't understand what "tag a workflow file" means here.

stone-z:
> Sorry but I'm lost. I think I don't understand what "tag a workflow file" means here.

Tagging a commit which modifies a workflow file

kopiczko:
> Tagging a commit which modifies a workflow file

That is not the case at all with release PRs.

stone-z:
Yeah I guess with the automated PR branch there won't be new commits, so we're good to go :+1: 

I'm looking at the create_release workflow and there's a lot to unpack, but I'll find a spot to tack this one onto that workflow.

pkg/gen/input/workflows/internal/file/create_release_branch.go
stone-z: Added this to create_release

stone-z:
Ok, I moved this to `create_release` and tested it in the workflow over on https://github.com/stone-z/test-gh-workflows

stone-z:
Pawel was able to get it working in bash, saving the trouble of embedding python. We've both tested it by now and it's a lot shorter :rocket:

pkg/gen/input/workflows/internal/file/create_release.go
```diff
@@ -204,6 +204,62 @@ jobs:
...
+            echo "Current tag is a new minor version"
+          else
+            echo "Current tag is not a new major or minor version. Nothing to do here."
+            exit 0
+          fi
+
+          release_branch="release-v${parent_major}.${parent_minor}.x"
+          echo "release_branch=$release_branch"
+
+          if git rev-parse --verify master $release_branch ; then
```
kopiczko, line 255: Last minute fix. I feel like we can remove this `if` altogether though. In the worst case scenario `git push` will fail and if it fails it's a good thing because we'll be notified that release branch head points at wrong commit.

```suggestion
          if git rev-parse --verify $release_branch ; then
```

stone-z: I think it's nice to have the explicit feedback, but we could make it a fail if the branch already exists. I don't think it's out of the question that we'd need to tag some branch besides master for a hotfix or x.x.x-0 release which could trigger this workflow. My only hesitation to make it fail is that it would then break downstream CI in such a case, which would be annoying in an already annoying situation. WDYT?

kopiczko: This job has no dependency, so the the action would be red, but release would be created anyway. I thought it would somewhat increase visibility if we have a conflict. If the branch already exists and points to the correct head then we'd only have "Already up to date" or something similar.

I leave the decision to you. We can iterate.

stone-z: Ok, I'll go with this and we can revisit if needed

PR #74, opened 2020-07-07
Dependabot

njuettner:
Generate new dependabot file

### DIFF
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -19,8 +19,7 @@ dependencies secure and up-to-date.
 E.g. run 'devctl gen dependabot --reviewers team/my-team' to automatically request
 a review from my-team.
 
-Omit reviewers flag and Dependabot will create a new PR without requesting
-a review.
+Omit reviewers flag and Dependabot will create a new PR without requesting a review.
 `
 )
 
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -11,7 +11,17 @@ import (
 
 const (
 	name        = "dependabot"
-	description = "Generates GitHub Dependabot Config."
+	description = `Generates GitHub Dependabot config (.github/dependabot.yml).
+
+This config is used by Dependabot to automatically creating pull requests to keep
+dependencies secure and up-to-date.
+
+Run 'devctl gen dependabot --reviewers team/my-team' to automatically request
+a review from my-team.
+
+Omit reviewers flag and Dependabot will create a new PR without requesting
+a review.
+`
 )
 
 type Config struct {
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -16,7 +16,7 @@ const (
 This config is used by Dependabot to automatically creating pull requests to keep
 dependencies secure and up-to-date.
 
-Run 'devctl gen dependabot --reviewers team/my-team' to automatically request
+E.g. run 'devctl gen dependabot --reviewers team/my-team' to automatically request
 a review from my-team.
 
 Omit reviewers flag and Dependabot will create a new PR without requesting
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -13,11 +13,9 @@ const (
 	name        = "dependabot"
 	description = `Generates GitHub Dependabot config (.github/dependabot.yml).
 
-This config is used by Dependabot to automatically creating pull requests to keep
-dependencies secure and up-to-date.
+This config is used by Dependabot to automatically creating pull requests to keep dependencies secure and up-to-date.
 
-E.g. run 'devctl gen dependabot --reviewers team/my-team' to automatically request
-a review from my-team.
+E.g. run 'devctl gen dependabot --reviewers team/my-team' to automatically request a review from my-team.
 
 Omit reviewers flag and Dependabot will create a new PR without requesting a review.
 `
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -1,0 +1,53 @@
+package dependabot
+
+import (
+	"io"
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/cobra"
+)
+
+const (
+	name        = "dependabot"
+	description = "Generates GitHub Dependabot Config."
+)
+
+type Config struct {
+	Logger micrologger.Logger
+	Stderr io.Writer
+	Stdout io.Writer
+}
+
+func New(config Config) (*cobra.Command, error) {
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+	if config.Stderr == nil {
+		config.Stderr = os.Stderr
+	}
+	if config.Stdout == nil {
+		config.Stdout = os.Stdout
+	}
+
+	f := &flag{}
+
+	r := &runner{
+		flag:   f,
+		logger: config.Logger,
+		stderr: config.Stderr,
+		stdout: config.Stdout,
+	}
+
+	c := &cobra.Command{
+		Use:   name,
+		Short: description,
+		Long:  description,
+		RunE:  r.Run,
+	}
+
+	f.Init(c)
+
+	return c, nil
+}
--- a/cmd/gen/dependabot/flag.go
+++ b/cmd/gen/dependabot/flag.go
@@ -10,6 +10,6 @@ type flag struct {
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
-	cmd.Flags().StringSliceVar(&f.Reviewers, "reviewers", []string{}, "Reviewers you want to assign automatically when Dependabot creates a PR, comma separated, e.g. giantswarm/team-firecracker.")
+	cmd.Flags().StringSliceVar(&f.Reviewers, "reviewers", []string{}, "Reviewers you want to assign automatically when Dependabot creates a PR, e.g. giantswarm/team-firecracker.")
 	cmd.Flags().BoolVar(&f.Daily, "daily", false, "Check for updates every day (default: weekly).")
 }
--- a/cmd/gen/dependabot/flag.go
+++ b/cmd/gen/dependabot/flag.go
@@ -6,8 +6,10 @@ import (
 
 type flag struct {
 	Reviewers []string
+	Daily     bool
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringSliceVar(&f.Reviewers, "reviewers", []string{}, "Reviewers you want to assign automatically when Dependabot creates a PR, comma separated, e.g. giantswarm/team-firecracker.")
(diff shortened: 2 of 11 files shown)


njuettner:
`devctl gen dependabot --reviewers njuettner,team/firecracker`, providing no reviewers omit the reviewer lines

```yaml
# DO NOT EDIT. Generated with:
#
#    devctl gen dependabot
#
version: 2
updates:
- package-ecosystem: gomod
  directory: "/"
  schedule:
    interval: weekly
    time: "04:00"
  open-pull-requests-limit: 10
  reviewers:
  - njuettner
  - team/firecracker
  ignore:
  - dependency-name: k8s.io/*
    versions:
    - ">=0.17.0"
```

njuettner:
Maybe we can optimise what should be in dependabot.yml, but as a first step it would be nice to generate this file via devctl.

pkg/gen/input/dependabot/internal/file/dependabot.go
```diff
@@ -0,0 +1,48 @@
...
+    time: "04:00"
+  open-pull-requests-limit: 10
+{{- if .Reviewers }}
+  reviewers:
+  {{- range $reviewer:= .Reviewers}} 
+  - {{$reviewer}} 
+   {{- end}}
+{{- end }}
+  ignore:
+  - dependency-name: k8s.io/*
```
stone-z, line 45: Has this rule been working somewhere? It might need to be `k8s.io/**` if not. I haven't received a response from the dependabot team about the exact supported syntax

pkg/gen/input/dependabot/internal/file/dependabot.go
```diff
@@ -0,0 +1,48 @@
...
+var createDependabotTemplate = `# DO NOT EDIT. Generated with:
+#
+#    devctl gen dependabot
+#
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: weekly
```
MarcelMue, line 35: This should be configurable IMO

pkg/gen/input/dependabot/internal/params/types.go
```diff
@@ -0,0 +1,9 @@
+package params
+
+type Params struct {
+	// Dir is the name of the directory where the files of the resource
+	// should be generated.
+	Dir string
+	// Reviewers is a set of people or teams who are assigned as reviewers
```
MarcelMue, line 7:
```suggestion
	// Reviewers is a set of people or teams who are assigned as reviewers.
```

cmd/gen/dependabot/command.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/cobra"
+)
+
+const (
+	name        = "dependabot"
+	description = "Generates GitHub Dependabot Config."
```
MarcelMue, line 14: I would love to have a long description with an example usage here.

cmd/gen/dependabot/flag.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+type flag struct {
+	Reviewers string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&f.Reviewers, "reviewers", "", "Reviewers you want to assign automatically when Dependabot creates a PR, comma separated.")
```
xh3b4sd, line 12:
```suggestion
	cmd.Flags().StringVar(&f.Reviewers, "reviewers", "", "Reviewers you want to assign automatically when Dependabot creates a PR, comma separated, e.g. giantswarm/team-firecracker.")
```

pkg/gen/gen.go
```diff
@@ -68,7 +68,7 @@ func execute(ctx context.Context, file input.Input) error {
 func isRegenerable(path string) bool {
```
xh3b4sd, line 68: I thought the `zz_generated.` prefix determines if a file can be re-generated. Why do we need to identify by file name?

pkg/gen/input/dependabot/internal/file/dependabot.go
xh3b4sd: In our operators I did not see k8s.io PRs. What we got though was `Bump sigs.k8s.io/cluster-api from 0.3.2 to 0.3.6`, which does also not work. So we can extend this expression to `*k8s.io/*` if that works.

stone-z: In the example you provided, it is a patch version bump so should have been done anyway. But if you wanted to exclude sigs dependencies I think you'd have to add them as separate ignore items because they don't all share the same versioning. Like in your example it still would have received a PR because it is not >=0.17.0

stone-z: So IMO it doesn't make sense to exclude that here by default

pkg/gen/gen.go
kopiczko: Some file names are conventional. Like `Makefile` or `dependabot.yml`. We still want to regenerate them.

cmd/gen/dependabot/flag.go
kopiczko: Let's also use `StringSliceVar`:

```suggestion
	cmd.Flags().StringSliceVar(&f.Reviewers, "reviewers", "", "Reviewers you want to assign automatically when Dependabot creates a PR, comma separated, e.g. giantswarm/team-firecracker.")
```

cmd/gen/dependabot/flag.go
```diff
@@ -0,0 +1,13 @@
+package dependabot
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type flag struct {
+	Reviewers string
```
kopiczko, line 8:
```suggestion
	Reviewers []string
```

cmd/gen/dependabot/runner.go
```diff
@@ -0,0 +1,64 @@
...
+			c.Reviewers = reviewerList(r.flag.Reviewers)
+		} else {
+			c.Reviewers = nil
+		}
+	}
+
+	dependabotInput, err := dependabot.New(c)
+	if err != nil {
+		return microerror.Mask(err)
+	}
```
kopiczko, line 49:
```suggestion
	var dependabotInput *dependabot.Dependabot
	{
		c := dependabot.Config{
			Reviewers: r.flag.Reviewers,
		}

		dependabotInput, err = dependabot.New(c)
		if err != nil {
			return microerror.Mask(err)
		}
	}
```

pkg/gen/input/dependabot/internal/file/dependabot.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func NewCreateDependabotInput(p params.Params) input.Input {
+	i := input.Input{
+		Path:         filepath.Join(p.Dir, "dependabot.yml"),
+		TemplateBody: createDependabotTemplate,
+		TemplateDelims: input.InputTemplateDelims{
+			Left:  "{{",
+			Right: "}}",
+		},
```
kopiczko, line 17: This is the default.

```suggestion
```

pkg/gen/input/dependabot/internal/file/dependabot.go
kopiczko: Why?

pkg/gen/input/dependabot/internal/file/dependabot.go
```diff
@@ -0,0 +1,48 @@
...
+  directory: "/"
+  schedule:
+    interval: weekly
+    time: "04:00"
+  open-pull-requests-limit: 10
+{{- if .Reviewers }}
+  reviewers:
+  {{- range $reviewer:= .Reviewers}} 
+  - {{$reviewer}} 
+   {{- end}}
```
kopiczko, line 42: That's my best suggestion so far:

```suggestion
  {{- end}}
```

pkg/gen/input/dependabot/internal/file/dependabot.go
```diff
@@ -0,0 +1,48 @@
...
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: weekly
+    time: "04:00"
+  open-pull-requests-limit: 10
+{{- if .Reviewers }}
+  reviewers:
+  {{- range $reviewer:= .Reviewers}} 
+  - {{$reviewer}}
```
kopiczko, line 41:
```suggestion
  {{- range $reviewer:= .Reviewers }} 
  - {{ $reviewer }} 
```

pkg/gen/input/dependabot/internal/file/dependabot.go
MarcelMue: Some repositories e.g. provider operators are getting updated quite often. So I would like to have daily for them.

kopiczko: Why not daily for everything then?

pkg/gen/gen.go
xh3b4sd: Makes sense. Good morning Tim. 😅

pkg/gen/input/dependabot/internal/file/dependabot.go
MarcelMue: Because some libraries are not so important - I would not want to have daily updates to `k8s-kvm` or something :D

kopiczko: I'm not convinced. Weekly seems to be good enough for me. But I won't block if that's what you want.

stone-z: Just my 2 cents:
Besides better velocity, daily actually provides a better dev experience because the PRs come one at a time, so they get integrated and tested individually. The overall time spent updating is shorter because the dev just approves one PR and is done for the day.

If using weekly, multiple PRs are opened at once which then need to be merged, but after merging each one the others then need time to rebase. Then there are a chunk of dependency changes. If one of them broke something, it is less clear which update it was.

If there are no updates on a given day, then there are no PRs anyway.

I'd say daily should be the default, but I'm fine configuring it.

kopiczko: I'm not convinced we should use different value in some places. Either daily or weekly works for me.

njuettner: I added a flag `--daily` to overwrite config from weekly schedule.

pkg/gen/input/dependabot/internal/file/dependabot.go
njuettner: Can I do something here to merge this PR? I can't follow the conversation.

xh3b4sd: You can keep it as is for now and move on.

njuettner:
I'll merge it, we can iterate over it 🚀 .

stone-z:
One last request - since the idea is to use this for everything, can we also include docker image dependencies by default? e.g.
```
- package-ecosystem: docker
  directory: "/"
  schedule:
    interval: daily
    time: "04:00"
  target-branch: master
  reviewers:
    - "giantswarm/team-whatever"
```

stone-z:
Haha, I was a second too slow. Ok, next iteration. Thanks for this, Nick

PR #78, opened 2020-07-08
autocomplete

njuettner:
Bump to the latest cobra version and enable autocompletion

Tests with oh-my-zsh:

Created a new dir "devctl" in `~/.oh-my-zsh/plugins/` and executed `devctl completion zsh > _devctl` inside this dir.
Finally added in `.zshrc` to load it:

```
plugins=(
  ...
  devctl
)
```
image


cmd/completion/command.go
```diff
@@ -0,0 +1,76 @@
...
+	name             = "completion [bash|zsh|fish|powershell]"
+	shortDescription = "Generate completion script."
+	longDescription  = `To load completions:
+	
+	Bash:
+	
+	$ source <(devctl completion bash)
+	
+	# To load completions for each session, execute once:
+	Linux:
```
kopiczko, line 22: Nit: shouldn't that also be commented?

PR #80, opened 2020-07-13
Support ecosystem for dependabot

njuettner:
Ecosystems flag will be mandatory, where you can specify what what package ecosystem you want to add.
This allows someone to include only dependabot ecosystems which are inside the repo.
E.g. update docker and go updates, or only update docker updates.

This also allows us to include more ecosystems like terraform, GitHub Actions, npm, ....

Current support:

- Docker
- Go

### DIFF
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -12,9 +12,9 @@ import (
 const (
 	name        = "dependabot"
 	description = "Generates GitHub Dependabot config for go and docker dependencies (.github/dependabot.yml)."
-	example     = `  devctl gen dependabot 
-  devctl gen dependabot --interval daily --reviewers giantswarm/team-firecracker
-  devctl gen dependabot --interval weekly --reviewers giantswarm/team-firecracker,njuettner`
+	example     = `  devctl gen dependabot --ecosystems go
+  devctl gen dependabot --ecosystems docker,go --interval daily --reviewers giantswarm/team-firecracker
+  devctl gen dependabot --ecosystems docker --interval weekly --reviewers giantswarm/team-firecracker,njuettner`
 )
 
 type Config struct {
--- a/cmd/gen/dependabot/command.go
+++ b/cmd/gen/dependabot/command.go
@@ -12,9 +12,9 @@ import (
 const (
 	name        = "dependabot"
 	description = "Generates GitHub Dependabot config for go and docker dependencies (.github/dependabot.yml)."
-	example     = `  devctl gen dependabot --ecosystems go
-  devctl gen dependabot --ecosystems docker,go --interval daily --reviewers giantswarm/team-firecracker
-  devctl gen dependabot --ecosystems docker --interval weekly --reviewers giantswarm/team-firecracker,njuettner`
+	example     = `  devctl gen dependabot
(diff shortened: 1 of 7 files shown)


cmd/gen/dependabot/command.go
```diff
@@ -12,9 +12,9 @@ import (
 const (
 	name        = "dependabot"
 	description = "Generates GitHub Dependabot config for go and docker dependencies (.github/dependabot.yml)."
-	example     = `  devctl gen dependabot 
-  devctl gen dependabot --interval daily --reviewers giantswarm/team-firecracker
-  devctl gen dependabot --interval weekly --reviewers giantswarm/team-firecracker,njuettner`
+	example     = `  devctl gen dependabot --ecosystems go
+  devctl gen dependabot --ecosystems docker,go --interval daily --reviewers giantswarm/team-firecracker
+  devctl gen dependabot --ecosystems docker --interval weekly --reviewers giantswarm/team-firecracker,njuettner`
```
kopiczko, line 17: I'm ok with that if you have a strong preferenc but maybe it would be simpler for the end user if the code would simply check `Dockerfile` existence?

njuettner: So you would drop out `Docker` as an exception but other package ecosystems would be ok like GH actions, terraform or npm? 

Would be ok to me I guess, I was not sure if we always use 'Dockerfile' as a file name or if something else might be also is used. I need some confirmation on that.

stone-z: To that point, it might be possible to make this optional and detect a Dockerfile / go.mod / yarn.lock / etc. and allow this to override the auto-detection

stone-z: If the repo is not using a standard name for these dependency files, then Dependabot can't find it anyway and the repo would have to apply a manual setting in the Dependabot config

njuettner: Ok sounds good, will change it a bit for the autodetection. Thanks both of you for the input 👍

stone-z: I couldn't find a list of the specific filenames they look for (they're spread between ruby modules) but here's a list of the languages at least https://dependabot.com/#languages

kopiczko: Let's maybe not go too crazy and detect only what we use and extend slowly.

cmd/gen/dependabot/flag.go
```diff
@@ -10,23 +11,56 @@ import (
...
-	Reviewers []string
+	Interval   string
+	Reviewers  []string
+	Ecosystems []string
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&f.Interval, flagInterval, "i", "weekly", "Check for daily, weekly or monthly updates (default: weekly).")
 	cmd.Flags().StringSliceVarP(&f.Reviewers, "reviewers", "r", []string{}, "Reviewers you want to assign automatically when Dependabot creates a PR, e.g. giantswarm/team-firecracker.")
+	cmd.Flags().StringSliceVarP(&f.Ecosystems, "ecosystems", "e", []string{}, "Ecosystem for each one package manager that you want GitHub Dependabot to monitor for new versions , e.g. go, docker")
```
stone-z, line 27: Could use `flagEcosystems` here for consistency

PR #81, opened 2020-07-17
Add lint-chart target to the generated Makefile

kopiczko:


pkg/gen/input/makefile/internal/file/makefile.go
```diff
@@ -132,6 +132,17 @@ build-docker: build-linux
...
 
+.PHONY: lint-chart
+## lint-chart: runs ct against the default chart
+lint-chart:
+	@echo "====> $@"
+	rm -rf /tmp/$(APPLICATION)-test
+	mkdir -p /tmp/$(APPLICATION)-test/helm
+	cp -a ./helm/$(APPLICATION) /tmp/$(APPLICATION)-test/helm/
+	architect helm template --dir /tmp/$(APPLICATION)-test/helm/$(APPLICATION)
+	docker run -it --rm -v /tmp/$(APPLICATION)-test:/wd --workdir=/wd --name ct quay.io/giantswarm/helm-chart-testing:v3.0.0-rc.1 ct lint --validate-maintainers=false --charts="helm/$(APPLICATION)"
```
njuettner, line 143: what about setting the image and version in the variable section in line 31?

kopiczko: Why do you think it's beneficial?

njuettner: In case of the image gets updated, was only a guess from my side that this happens at some point 😄. But no blocker from my side.

kopiczko: I still want to drill it because I have my reasons to keep it here and I may change my mind if understand you better. Why do you think it's better to have it up in the line 31 section if the image gets updated?

My thinking was that if we remove the target we won't have any leftover variables. This is my rule of thumb when using the value once.

njuettner: The docker image there is not easy to see, so we might forget to update when needed. I thought it would be easier to see when it's in the vars section, so you don't have to scan through the lines.

kopiczko, line 143: How about this?

```suggestion
.PHONY: lint-chart
## lint-chart: runs ct against the default chart
lint-chart: IMAGE := quay.io/giantswarm/helm-chart-testing:v3.0.0-rc.1
lint-chart:
	@echo "====> $@"
	rm -rf /tmp/$(APPLICATION)-test
	mkdir -p /tmp/$(APPLICATION)-test/helm
	cp -a ./helm/$(APPLICATION) /tmp/$(APPLICATION)-test/helm/
	architect helm template --dir /tmp/$(APPLICATION)-test/helm/$(APPLICATION)
	docker run -it --rm -v /tmp/$(APPLICATION)-test:/wd --workdir=/wd --name ct $(IMAGE) ct lint --validate-maintainers=false --charts="helm/$(APPLICATION)"
```

njuettner: nice 👍 , lgtm

PR #87, opened 2020-08-06
Update apiextensions to save the releaseOperatorDeploy field

tfussell:
The `releaseOperatorDeploy` was being lost when unmarshalling, editing, and marshalling the Release in `devctl release create` because it was using a version of `apiextensions` without that field.


tfussell:
Dependency version update only, going with this.

PR #89, opened 2020-08-11
Skip reference version validations in release workflow

corest:


pkg/gen/input/workflows/internal/file/create_release.go
```diff
@@ -82,6 +82,13 @@ jobs:
           fi
           echo "path=\"$path\""
           echo "::set-output name=path::${path}"
+      - name: Check if reference version
+        id: ref_version
+        run: |
+          version=$(echo $title | cut -d ' ' -f 2)
+          if [[ "${version}" =~ ^[0-9]+.[0-9]+.[0-9]+-[0-9]+$ ]]; then
+            echo ::set-output name=refversion::true
+          fi
```
kopiczko, line 91:
```suggestion
          refversion=false
          if [[ "${{ steps.get_version.outputs.version }}" =~ ^[0-9]+.[0-9]+.[0-9]+-[0-9]+$ ]]; then
            refversion=true
          fi
          echo "refversion =\"$refversion\""
          echo ::set-output name=refversion::true
```

PR #93, opened 2020-08-18
Update go modules to prepare for K8s 1.18

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/12716

Upgrading k8sclient to v3. Once this is merged I'll use the Makefile to upgrade to 1.18.


rossf7:
Going with a ping.

PR #96, opened 2020-08-28
Add GitHub actions to dependabot workflow

stone-z:
Adds the `github-actions` ecosystem and auto-detection to generated dependabot configs


MarcelMue (approved):

kopiczko:
What does it upgrade? Used actions?

stone-z:
> What does it upgrade? Used actions?

Yeah exactly. So the `create-release` workflows shouldn't really be affected (although it might eventually bump their `runs-on` version, not sure), but any external actions (like `gitleaks` or the aws stuff on `installations`) will be bumped.

PR #98, opened 2020-09-11
Use the full path for setting the gitSHA in Readme

ubergesundheit:
This PR corrects the `go install -ldflags..` command in readme to correctly include the gitSHA in the installed binary.


MarcelMue (approved):
LGTM if tested

PR #106, opened 2020-09-24
Update actions/cache and actions/upload-artifact to v2.1.1

kopiczko:


axbarsan (approved):
LGTM if tested

kopiczko:
I'm coming from:

- https://github.com/giantswarm/operatorkit/pull/456
- https://github.com/giantswarm/operatorkit/pull/457

So I consider that tested.

kopiczko:
We have 1.0.0 already. I'll sort that later.

MarcelMue (approved):

njuettner (approved):

xh3b4sd:
This is a little odd and it looks like how we use Github actions is not quite right. We will always stay behind with our funny golang string template in `devctl` and dependabot will always be ahead updating actions in our workflows. Are there other options we have to manage actions and workflows? I was thinking that we could maybe checkout how to host our actions in a proper way so that they are consumable via the Github marketplace. I am not sure how it works but I would hope that dependabot would be able to update our action repositories as well. Then a chain of updates would get triggered and we do not have to play this catchup ball game.

MarcelMue:
The problem is currently not easy to solve - unless we want to reimplement every upstream action ourselves again. I assume GitHub is working on some way to reuse workflows for now you can not use actions in actions:

https://github.community/t/can-i-use-action-as-a-full-job-to-encapsulate-all-ci-process/17238

kopiczko:
It's good enough for me for now. Ideally dependabot could update Go templates here, but I don't see how we can achieve that now.

PR #117, opened 2020-10-12
Add executed command as comment for release creation

MarcelMue:
This has annoyed me one time too many now. We always lose this execution. This adds it as a comment to the release YAML.

AFAIK @tfussell wanted to do this?

### DIFF
--- a/cmd/release/create/runner.go
+++ b/cmd/release/create/runner.go
@@ -2,7 +2,10 @@ package create
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"os"
+	"strings"
 
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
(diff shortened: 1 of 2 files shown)


MarcelMue:
Looks like this for example:
```
# Generated with:
# devctl release create --name v12.10.0 --base v12.2.0 --releases . --provider kvm --component containerlinux@2512.4.0 --app cert-exporter@1.2.4-c2858702bf1c5dd500892c203677169ec973a983 --overwrite true
apiVersion: release.giantswarm.io/v1alpha1
kind: Release
metadata:
  creationTimestamp: null
  name: v12.10.0
...
```

stone-z (approved):
LGTM, saw it in action on your release today

tfussell (approved):
Sweet. I wanted to have a more domain-specific representation of the command that was used, but this already provides a lot of value.

PR #134, opened 2020-10-15
Use giantswarm/install-tools-action in cli flavour generated workflows

kopiczko:
I forgot about that.


MarcelMue (approved):
How dare you forget something :trollface:

PR #158, opened 2020-10-28
Simplify gen flavours

kopiczko:
As discussed here https://gigantic.slack.com/archives/C014ZAZLEMT/p1603820086002100


fiunchinho (changes requested):
Can you improve the docs for the command? Just what you wrote me in slack would be an improvement.

fiunchinho (approved):

PR #166, opened 2020-11-20
Remove repo list command

kopiczko:
It was listing dependencies of a repo based on Gopkg.toml files used by
godep. We don't use godep anymore. It was supposed to be used for
checking where dependencies need to be updated. Since we have dependabot
we don't need that as well.


MarcelMue (approved):
Fine with me

mig4 (approved):

PR #170, opened 2020-11-24
Add `gen enum` command

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/14514


kopiczko:
Won't pursue that for now.

PR #177, opened 2020-11-26
Refactor gen dependabot ; Rename ecosystem go to gomod

kopiczko:


cmd/gen/dependabot/flag.go
```diff
@@ -35,23 +35,15 @@ func (f *flag) Validate() error {
 
 	// in case ecosystem was not set specifically, we autodetect files
 	if len(f.Ecosystems) == 0 {
-		for _, e := range gen.AllowedEcosystems() {
```
kopiczko, line 38: This was the first strange thing.

pkg/gen/input/dependabot/internal/file/dependabot.go
```diff
@@ -41,38 +42,11 @@ updates:
...
-    time: "04:00"
-  target-branch: master
-{{- if $reviewers }}
-  reviewers:
-  {{- range $reviewer := $reviewers }}
-  - {{ $reviewer }}
-  {{- end}}
-{{- end }}
-{{- end }}
 {{- end }}
```
kopiczko, line 51: There was a lot of repetition. The for loop is better utilised now.

pkg/gen/dependabot_ecosystem.go
```diff
@@ -9,14 +9,14 @@ import (
 const (
 	EcosystemDocker        Ecosystem = "docker"
 	EcosystemGithubActions Ecosystem = "github-actions"
-	EcosystemGo            Ecosystem = "go"
+	EcosystemGomod         Ecosystem = "gomod"
```
kopiczko, line 12: I renamed that to match the real name. That way it will be easy to add new ecosystems.

PR #181, opened 2020-12-04
Set version flag in generated Makefiles

rossf7:


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Generate main Makefile including `*.mk` files to allow custom Makefiles.
 - Pretty print errors.
 - Print devctl version in generated files headers.
+- Set version flag in generated Go Makefiles.
```
kopiczko, line 16: I'd call it Fixed

pkg/gen/input/makefile/internal/file/makefile_gen_go_mk.go
```diff
@@ -35,8 +35,9 @@ SOURCES        := $(shell find . -name '*.go')
 VERSION        := $(shell architect project version)
 LDFLAGS        ?= -w -linkmode 'auto' -extldflags '-static' \
   -X '$(shell go list .)/pkg/project.buildTimestamp=${BUILDTIMESTAMP}' \
-  -X '$(shell go list .)/pkg/project.gitSHA=${GITSHA1}'
-.DEFAULT_GOAL := build
+  -X '$(shell go list .)/pkg/project.gitSHA=${GITSHA1}' \
+  -X '$(shell go list .)/pkg/project.version=${VERSION}'
+  .DEFAULT_GOAL := build
```
kopiczko, line 40: This is a Makefile spaces are important here :) 

```suggestion
.DEFAULT_GOAL := build
```

kopiczko (approved):

pkg/gen/input/makefile/internal/file/makefile_gen_go_mk.go
```diff
@@ -35,8 +35,9 @@ SOURCES        := $(shell find . -name '*.go')
 VERSION        := $(shell architect project version)
 LDFLAGS        ?= -w -linkmode 'auto' -extldflags '-static' \
   -X '$(shell go list .)/pkg/project.buildTimestamp=${BUILDTIMESTAMP}' \
-  -X '$(shell go list .)/pkg/project.gitSHA=${GITSHA1}'
-.DEFAULT_GOAL := build
+  -X '$(shell go list .)/pkg/project.gitSHA=${GITSHA1}' \
+  -X '$(shell go list .)/pkg/project.version=${VERSION}'
```
kopiczko, line 39: But don't we want this to be set in `version.go`? We don't want this to be overwritten at the compile time.

rossf7:
Thanks Pawel I'll get both those fixed. I meant to click draft 😄 

This is because we added it to apptestctl in https://github.com/giantswarm/apptestctl/pull/58 but it's now being removed when the files are aligned.

kopiczko:
> This is because we added it to apptestctl in [giantswarm/apptestctl#58](https://github.com/giantswarm/apptestctl/pull/58) but it's now being removed when the files are aligned.

But why is it different there?

pkg/gen/input/makefile/internal/file/makefile_gen_go_mk.go
rossf7: The problem is when the binary is downloaded as a GitHub release. 
In that case there is no version. 

```
apptestctl version
Git Commit:     n/a
Go Version:     go1.15.3
OS / Arch:      darwin / amd64
Source:         https://github.com/giantswarm/apptestctl
```

This is what was fixed in https://github.com/giantswarm/apptestctl/pull/58

rossf7:
We don't need the version flag in the Makefile.

PR #183, opened 2020-12-04
Add npm and pip ecosystems

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/14728

~~There is still some work left to use it with giantswarm/github.~~ Autodetection based on files existence in place.

### DIFF
--- a/cmd/gen/dependabot/flag.go
+++ b/cmd/gen/dependabot/flag.go
@@ -44,6 +44,12 @@ func (f *flag) Validate() error {
 		if exists("go.mod") && exists("go.sum") {
 			f.Ecosystems = append(f.Ecosystems, gen.EcosystemGomod.String())
 		}
+		if exists("package.json") {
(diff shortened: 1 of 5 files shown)


cmd/gen/dependabot/flag.go
```diff
@@ -44,6 +44,12 @@ func (f *flag) Validate() error {
 		if exists("go.mod") && exists("go.sum") {
 			f.Ecosystems = append(f.Ecosystems, gen.EcosystemGomod.String())
 		}
+		if exists("package.json") {
+			f.Ecosystems = append(f.Ecosystems, gen.EcosystemNPM.String())
+		}
+		if exists("setup.py") {
+			f.Ecosystems = append(f.Ecosystems, gen.EcosystemPIP.String())
+		}
```
kopiczko, line 52: @axbarsan @glitchcrab if you think we should detect on different files LMK.

axbarsan (approved):
Noice

glitchcrab (approved):
this approach makes sense :+1:

PR #200, opened 2021-01-14
Fix module inferring in Makefile for Go

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/14992

### DIFF
--- a/pkg/gen/input/makefile/internal/file/makefile_gen_go_mk.go
+++ b/pkg/gen/input/makefile/internal/file/makefile_gen_go_mk.go
@@ -27,15 +27,16 @@ PACKAGE_DIR    := ./bin-dist
 
 {{- end }}
 
-APPLICATION    := $(shell go list . | cut -d '/' -f 3)
+APPLICATION    := $(shell go list -m | cut -d '/' -f 3)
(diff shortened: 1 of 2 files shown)


tomahawk28 (approved):
<3

pkg/gen/input/makefile/internal/file/makefile_gen_go_mk.go
```diff
@@ -27,15 +27,16 @@ PACKAGE_DIR    := ./bin-dist
 
 {{- end }}
 
-APPLICATION    := $(shell go list . | cut -d '/' -f 3)
+APPLICATION    := $(shell go list -m | cut -d '/' -f 3)
 BUILDTIMESTAMP := $(shell date -u '+%FT%TZ')
 GITSHA1        := $(shell git rev-parse --verify HEAD)
+LOCAL_IMPORT   := $(shell go list -m | cut -d '/' -f 1-3)
```
tfussell, line 33: I think you can use plain `go list -m` here (i.e. without `cut`). `goimports -local` matches a prefix so `github.com/giantswarm/apiextensions` or `github.com/giantswarm/apiextensions/v3` should both be fine, and the second one is simpler.

tfussell (approved):
LGTM.

PR #212, opened 2021-01-29
create release on main branch

kubasobon:
How did I miss this :(


pkg/gen/input/workflows/internal/file/create_release.go
```diff
@@ -28,6 +28,7 @@ on:
   push:
     branches:
       - 'legacy'
+	  - 'main'
```
MarcelMue, line 31: This just made all workflows fail. Please no tabs no more :heart:

kubasobon: :facepalm: On it

PR #227, opened 2021-02-24
Add arm64 builds

ubergesundheit:
This PR adds targets for `darwin-arm64` and `linux-arm64` to the generated Makefile for `--language go`.

Other changes

- Bump the used `architect` version in generated workflows
- Add `darwin-arm64` and `linux-arm64` targets to build matrix in workflows

Disclaimer: I have no idea how Makefiles work, I just copied around stuff until it worked on my machine:tm:

I would say, treat this PR with low priority..

### DIFF
--- a/cmd/gen/workflows/flag.go
+++ b/cmd/gen/workflows/flag.go
@@ -13,21 +13,28 @@ import (
 const (
 	flagCheckSecrets = "check-secrets"
 	flagFlavour      = "flavour"
+	flagArchitecture = "architecture"
 )
 
 type flag struct {
-	CheckSecrets bool
-	Flavours     gen.FlavourSlice
+	CheckSecrets  bool
+	Flavours      gen.FlavourSlice
+	Architectures gen.ArchitectureSlice
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().VarP(gen.NewFlavourSliceFlagValue(&f.Flavours, gen.FlavourSlice{}), flagFlavour, "f", fmt.Sprintf(`The type of project that you want to generate the workflows for. Possible values: <%s>`, strings.Join(gen.AllFlavours(), "|")))
 	cmd.Flags().BoolVar(&f.CheckSecrets, flagCheckSecrets, true, "If true, also generate a secret-scanning workflow. Possible values: true (default), false.")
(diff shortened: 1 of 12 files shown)


MarcelMue (approved):

ubergesundheit:
Do you have any advice on how to go forward testing this without too much disruption? Make a beta release of devctl to use with a branch of giantswarm/github?

ubergesundheit:
Ping @sig-ux just for awareness

I'm planning to revisit this PR in the near future

kopiczko:
@ubergesundheit what I usually do is `make install` and then `devctl gen` against my test repository. It's here if you need inspiration: https://github.com/kopiczko/test-gh-workflows.

kopiczko (approved):

cmd/gen/workflows/flag.go
```diff
@@ -13,22 +13,29 @@ import (
...
+	cmd.Flags().VarP(gen.NewArchitectureSliceFlagValue(&f.Architectures, gen.ArchitectureSlice{gen.ArchitectureDarwin, gen.ArchitectureLinux}), flagArchitecture, "a", fmt.Sprintf(`The architectures to build release artifacts for through workflows. Possible values: <%s>`, strings.Join(gen.AllArchitectures(), "|")))
 }
 
 func (f *flag) Validate() error {
 	if len(f.Flavours) == 0 {
 		return microerror.Maskf(invalidFlagError, "--%s must be one of: %s", flagFlavour, strings.Join(gen.AllFlavours(), ", "))
 	}
 
+	if len(f.Architectures) == 0 {
+		return microerror.Maskf(invalidFlagError, "--%s must contain only values of: %s", flagArchitecture, strings.Join(gen.AllArchitectures(), ", "))
```
kopiczko, line 37: Why not all of them always?

ubergesundheit:
Tested in https://github.com/ubergesundheit/test-gh-workflows/actions/runs/667938810 which created binaries for https://github.com/ubergesundheit/test-gh-workflows/releases/tag/v15.0.4

cmd/gen/workflows/flag.go
ubergesundheit: I changed my mind. 2754453 enables all architectures by default :)

kopiczko: I don't see why someone would care to disable any of the architectures. This is only for the binaries attached to the release..

kopiczko: Just came to mind: so far we were _always_ releasing for darwin and linux. If we _always_ release for darwin then we should _always_ release both amd64 and arm64. Then we can get rid of the flag and the enum and simplify the code here.

ubergesundheit: I’ll create another PR. 👍

PR #238, opened 2021-04-26
Don't crash if the container linux version is not present in changelog

whites11:
If the container linux version being set for a release is not present in the changelog upstream devctl crashes:

```
Error: Unexpected end of JSON input
        /home/whites/workspace/devctl/pkg/release/changelog/containerlinux.go:27
        /home/whites/workspace/devctl/pkg/release/changelog/changelog.go:246
        /home/whites/workspace/devctl/pkg/release/release_notes.go:57
        /home/whites/workspace/devctl/pkg/release/create.go:102
        /home/whites/workspace/devctl/cmd/release/create/runner.go:44
        /home/whites/workspace/devctl/cmd/release/create/runner.go:34
        /home/whites/workspace/devctl/main.go:50
```

This PR aims at avoiding the crash and setting an empty changelog in the release readme instead

### DIFF
--- a/pkg/release/changelog/containerlinux.go
+++ b/pkg/release/changelog/containerlinux.go
@@ -2,6 +2,7 @@ package changelog
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/giantswarm/microerror"
 )
@@ -16,8 +17,14 @@ func parseContainerLinuxChangelog(body []byte, componentVersion string) (string,
 	if err != nil {
 		return "", microerror.Mask(err)
 	}
+
+	b, ok := releases[componentVersion]
+	if !ok {
+		return fmt.Sprintf("Containerlinux release %q was not found in the changelog", componentVersion), nil
+	}
+
 	var release containerlinuxRelease
-	err = json.Unmarshal(releases[componentVersion], &release)
+	err = json.Unmarshal(b, &release)
 	if err != nil {
 		return "", microerror.Mask(err)
 	}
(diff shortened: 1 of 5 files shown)


kopiczko (approved):
That works for me but I'm not sure if that's a desired behaviour. I requested a review from @giantswarm/team-ludacris just in case.

pkg/release/changelog/containerlinux.go
```diff
@@ -16,8 +17,14 @@ func parseContainerLinuxChangelog(body []byte, componentVersion string) (string,
 	if err != nil {
 		return "", microerror.Mask(err)
 	}
+
+	b, ok := releases[componentVersion]
+	if !ok {
+		return fmt.Sprintf("Containerlinux release %q was not found in the changelog", componentVersion), nil
```
MarcelMue, line 23: IMO this should return an error in that case. The changelog for containerlinux should never be empty.

whites11: I don't agree.
If for example we are testing a non stable version of containerlinux the `devctl release` command fails and you have to generate the release templates yourself.
The behaviour is similar when using an unreleased version of an operator for example. The changelog being set is "Not found".
I don't see a reason why it should behave differently with Containerlinux

MarcelMue: Is something else wrong then? AFAIK every flatcar release - even alpha channel - has release notes:
https://kinvolk.io/flatcar-container-linux/releases-json/releases.json

whites11: the url was changed (https://github.com/giantswarm/devctl/pull/240)

MarcelMue: So is this still needed with #240 in place?

whites11: If you ask me: yes.
But you are clearly against it so I will close it because I have no strong opinions anyway.

PR #245, opened 2021-05-04
Fix self-update cache

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/16949

This blows up giantswarm/github: https://github.com/giantswarm/github/runs/2500122692

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -1,7 +1,9 @@
 package updater
 
 import (
+	"fmt"
 	"net/url"
+	"os"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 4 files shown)


pkg/updater/updater.go
```diff
@@ -138,7 +140,10 @@ func (u *Updater) getLatestVersion() (semver.Version, error) {
 		u.cache.LatestVersion = latestVersion.Version.String()
 		// We ignore the error, because if saving fails, we'll just
 		// fetch the version again next time.
-		_ = u.cache.Persist(u.cacheDir)
+		err := u.cache.Persist(u.cacheDir)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "WARNING: Failed to persist the cache with error: %s", err)
+		}
```
kopiczko, line 146: That shouldn't interrupt the work but at least gives a heads up that the caching is broken and a chance to fix it if so.

axbarsan (approved):
Nice 👍🏻 

Also, I think we can use the `DEVCTL_UNSAFE_FORCE_VERSION` env variable to disable checking for updates altogether, since we're explicitly installing the latest version.

kopiczko:
> Also, I think we can use the `DEVCTL_UNSAFE_FORCE_VERSION` env variable to disable checking for updates altogether, since we're explicitly installing the latest version.

Agreed.

PR #252, opened 2021-05-18
Use embed for templates

marians:
For cleaner, easier to maintain code, this PR moves templates into separate files.


kopiczko (approved):
Nice, thank you!

PR #261, opened 2021-06-21
Revert "Align files"

MarcelMue:
Reverts giantswarm/devctl#248


MarcelMue:
cc @stone-z reverting this as it removed the create release workflow

PR #262, opened 2021-06-21
 Fix templating of release workflow

MarcelMue:


MarcelMue:
@marians Is this expected to work? Seems like there was no changelog previously so I rolled out a broken release of devctl.

stone-z (approved):

anvddriesch (approved):

PR #281, opened 2021-08-09
Add renonvate command

tomahawk28:
Adding `devctl gen renovate` command to add the renovate configuration file. 

Toward https://github.com/giantswarm/giantswarm/issues/17522

### Checklist

### DIFF
--- a/cmd/gen/renovate/command.go
+++ b/cmd/gen/renovate/command.go
@@ -1,0 +1,1 @@
+package renovate
--- a/cmd/gen/renovate/command.go
+++ b/cmd/gen/renovate/command.go
@@ -1,1 +1,56 @@
 package renovate
+
+import (
+	"io"
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/cobra"
+)
+
+const (
+	name        = "renovate"
+	description = "Generates Renovate config for go and docker dependencies (renovate.yml)."
+	example     = `  devctl gen renovate 
+  devctl gen renovate --interval "after 9am on thursday" --reviewers giantswarm/team-firecracker`
+)
+
+type Config struct {
+	Logger micrologger.Logger
+	Stderr io.Writer
+	Stdout io.Writer
+}
+
+func New(config Config) (*cobra.Command, error) {
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+	if config.Stderr == nil {
+		config.Stderr = os.Stderr
+	}
+	if config.Stdout == nil {
+		config.Stdout = os.Stdout
+	}
+
+	f := &flag{}
+
+	r := &runner{
+		flag:   f,
+		logger: config.Logger,
+		stderr: config.Stderr,
+		stdout: config.Stdout,
+	}
+
+	c := &cobra.Command{
+		Use:     name,
+		Short:   description,
+		Long:    description,
+		Example: example,
+		RunE:    r.Run,
+	}
+
+	f.Init(c)
+
+	return c, nil
+}
--- a/cmd/gen/renovate/flag.go
+++ b/cmd/gen/renovate/flag.go
@@ -1,1 +1,37 @@
 package renovate
+
+import (
+	"github.com/giantswarm/microerror"
+	"github.com/spf13/cobra"
+)
+
+const (
+	flagInterval = "interval"
+	flagLanguage = "language"
+	flagReviewer = "reviewer"
+)
+
+type flag struct {
+	Interval string
+	Language string
+	Reviewer string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&f.Interval, flagInterval, "i", "weekly", "Check for daily, weekly or monthly updates.")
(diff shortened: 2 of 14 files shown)


cmd/gen/renovate/flag.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type flag struct {
+	Interval string
+	Language string
+	Reviewer string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&f.Interval, flagInterval, "i", "after 9am on thursday", "Check for daily, weekly or monthly updates.")
```
tomahawk28, line 21: Give out default interval value as `after 9am on thursday`

pkg/gen/input/renovate/internal/file/renovate.json.template
```diff
@@ -0,0 +1,36 @@
...
+      "allowedVersions": "< 1.21.0"
+    },
+    {
+      "matchPackageNames": ["sigs.k8s.io/controller-runtime"],
+      "allowedVersions": "< 0.7.0"
+    }
+  ],
+  "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],
+  {{- end }}
+  "reviewers": [ {{ $reviewer | printf "team:%s" | printf "%q" }} ],
```
tomahawk28, line 33: I need to put `team:` as a prefix if this is team. https://docs.renovatebot.com/configuration-options/#reviewers

cmd/gen/renovate/runner.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	f, err := filepath.Abs("./.github/dependabot.yml")
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = os.Remove(f)
```
tomahawk28, line 73: Delete `dependabot.yaml` to avoid having duplicated PRs.

cmd/gen/renovate/command.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/cobra"
+)
+
+const (
+	name        = "renovate"
+	description = "Generates Renovate config for go and docker dependencies (renovate.yml)."
```
rossf7, line 14: Direction is nice and we've tested the Go support a fair bit. My main concern is about the repos using Python for tests.

We use pip with dependabot like here

https://github.com/giantswarm/nginx-ingress-controller-app/blob/28548c426b944a63c97302ecf8148bd4c77ea862/.github/dependabot.yml#L1

Have you had a chance to test the Python support? Maybe Halo can helm with that.

pkg/gen/input/renovate/internal/file/renovate.json.template
```diff
@@ -0,0 +1,36 @@
...
+  "ignoreDeps": ["zricethezav/gitleaks-action"],
```
rossf7, line 34: Nice 👍

cmd/gen/renovate/command.go
tomahawk28: renovate also support pip dependency upgrade. https://docs.renovatebot.com/python/ 
But I haven't had any chance to test this in python. Actually, I'm thinking that using renovate only for the team who opt-in by checking https://github.com/giantswarm/github/blob/master/repositories/meta.yaml. We can first use renovate from go project and slowly roll out to the different projects who use a different language other than go.  WDYT?

rossf7: > Actually, I'm thinking that using renovate only for the team who opt-in by checking https://github.com/giantswarm/github/blob/master/repositories/meta.yaml. We can first use renovate from go project and slowly roll out to the different projects who use a different language other than go.

Yes, good idea. Better to make it opt-in and then teams can migrate at their own pace and try out support for other languages.

tomahawk28:
@giantswarm/team-biscuit Can someone review this?

pkg/gen/input/renovate/internal/file/renovate.json.template
```diff
@@ -0,0 +1,36 @@
...
+      "excludePackagePatterns": [".*giantswarm.*" ,"^k8s.io", "^sigs.k8s.io"],
+      "groupName": "etc modules"
+    },
+    {
+      "matchPackagePatterns": ["^k8s.io"],
+      "allowedVersions": "< 1.21.0"
+    },
+    {
+      "matchPackageNames": ["sigs.k8s.io/controller-runtime"],
+      "allowedVersions": "< 0.7.0"
```
rossf7, line 28: Is there a way to have additional rules but just for a specific repo?

It would be nice to have `helm.sh/helm/v3 < 3.6.0`. We only need that in helmclient but we might have similar requirements in other repos. WDYT?

rossf7 (approved):
LGTM when biscuit is happy

pkg/gen/input/renovate/internal/file/renovate.json.template
tomahawk28: We can extend github meta data for that. What do you think to extend https://github.com/giantswarm/github/tree/master/repositories/override ? 
For example, we can merge `override/chart-operator/renovate.json` with the default `renovate.json`

rossf7: Nice, I'd forgotten about the overrides. I think that will work well.

PR #284, opened 2021-08-12
Fixing renovate template issues

tomahawk28:
1. Down the version limit of k8s to 0.21.0 
2. `--reviewer` is not required for `devctl gen renovate` anymore.  

### Checklist

### DIFF
--- a/pkg/gen/input/renovate/internal/file/renovate.json.template
+++ b/pkg/gen/input/renovate/internal/file/renovate.json.template
@@ -4,6 +4,9 @@
 {
   "extends": [
     "config:base"
+    {{- if $reviewer }},
+    ":reviewer({{ $reviewer | printf "team:%s" }})"
+    {{- end }}
   ],
   {{- if eq $language "go" }}
   "packageRules": [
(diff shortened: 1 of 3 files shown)


pkg/gen/input/renovate/internal/file/renovate.json.template
```diff
@@ -4,6 +4,9 @@
 {
   "extends": [
     "config:base"
+    {{- if $reviewer }},
+    ":reviewer({{ $reviewer | printf "team:%s" }})"
```
tomahawk28, line 8: I think using extends for reviewers is more cleaner.

rossf7: How is reviewer set if there is no flag?

rossf7 (approved):
Fine with me. Just a query on how extends works.

pkg/gen/input/renovate/internal/file/renovate.json.template
tomahawk28: It would assign CODEOWNER as a reviewer automatically. I think it makes sense to loose the restriction on the reviewer flag when it has been not working properly from the renovate side.

tomahawk28:
Just a small fix :) I'm going without approval from team-biscuit.

PR #311, opened 2021-12-20
fix Get version bug

TheoBrigitte:
With some commit messages the `Get version` job fails with

```
Run title="$(echo "disable capi controller on kvm (#782)
/home/runner/work/_temp/8532b5f1-4835-41e4-9331-7127a6116541.sh: line 1: unexpected EOF while looking for matching `)'
Error: Process completed with exit code 2.
```

see: https://github.com/giantswarm/prometheus-meta-operator/runs/4583563350?check_suite_focus=true

I also went ahead and protected aligned variables protection with quotes and curly braces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):

TheoBrigitte:
Tested with https://github.com/giantswarm/prometheus-meta-operator/pull/783 :heavy_check_mark:

PR #390, opened 2022-04-01
Automatically bump go module version when releasing a major release

fiunchinho:
It happened to me many times that when releasing a new major version of a library or operator I forget to bump the go module version defined in the `go.mod` file, along with all the imports in the code. This is a manual and time consuming task that could be automated.

I found a couple of third party tools that do this task, so I figured we could integrate it in our release workflows so that this is done automatically when releasing. I ended up choosing https://github.com/marwan-at-work/mod.

I changed the workflow so that the third party tool is only installed and used when we are releasing a new major release and the `go.mod` file exists, so that we don't mess with apps or other flavours.

### Checklist

- [X] Update changelog in CHANGELOG.md.


pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
```diff
@@ -126,6 +137,12 @@ jobs:
       - name: Prepare release changes
         run: |
           architect prepare-release ${{ env.architect_flags }} --version "${{ needs.gather_facts.outputs.version }}"
+      - name: Bump go module defined in go.mod if needed
+        run: |
+          if [ "${{ needs.gather_facts.outputs.is_major }}" = true ] && test -f "go.mod"; then
+            go install github.com/marwan-at-work/mod/cmd/mod@v0.4.1
```
kopiczko, line 143: Isn't `go install` deprecated? Don't they provide a binary?

fiunchinho: Nope AFAIK https://github.com/marwan-at-work/mod/releases

kopiczko, line 143: I was wrong:
```
$ docker run  golang go version
go version go1.18 linux/amd64
```

```
$ docker run  golang sh -c 'GO111MODULE=on go get github.com/marwan-at-work/mod/cmd/mod@v0.4.1'
go: go.mod file not found in current directory or any parent directory.
        'go get' is no longer supported outside a module.
        To build and install a command, use 'go install' with a version,
        like 'go install example.com/cmd@latest'
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
```

pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
```diff
@@ -114,6 +122,9 @@ jobs:
     env:
       architect_flags: "--organisation ${{ github.repository_owner }} --project ${{ needs.gather_facts.outputs.repo_name }}"
     steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '=1.17.0'
```
kopiczko, line 127: Why not 18?

kopiczko (approved):

pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
fiunchinho: We are using `1.17` in architect and most of our repositories specify 1.17 on their `go.mod` files, so I figure we would want to upgrade everything to 1.18 at the same time.

TheoBrigitte (approved):
I like this, nice job :+1:

PR #395, opened 2022-04-06
Git: Synchronize .gitignore with kubectl-gs and opsctl.

Gacko:
- Ignore bin-dist & TODO.
- Ignore macOS & IDE specific files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,9 @@
+# Project
 devctl*
+bin-dist/
+TODO
+
+# OS & IDE
+.DS_Store
+.idea/
+.vscode/


AverageMarcus (approved):

.gitignore
```diff
@@ -1 +1,9 @@
+# Project
 devctl*
+bin-dist/
+TODO
+
+# OS & IDE
+.DS_Store
+.idea/
+.vscode/
```
kopiczko, line 9: IMO you should have this stuff in your global gitignore

Gacko: I already thought about something like this, but someone else added them to other repositories, too. So: Is there a doc about this? At least I didn't find one, also took a look inside giantswarm/fmt.

kopiczko: Not really, not a big deal.

PR #379, opened 2022-03-30
Add GitHub workflow for checking missing values schema changes

ubergesundheit:
This PR adds the already existing GitHub workflow to be templated using devctl.

I've changed the original to not use any external GitHub actions

This PR builds upon the excellent work of @glitchcrab done in https://github.com/giantswarm/template-app/pull/14

### Checklist

- [x] Update changelog in CHANGELOG.md.

PR #405, opened 2022-04-14
Add windows-amd64 to platforms for CLI builds

marians:
Towards https://github.com/giantswarm/roadmap/issues/333

This adds building of `windows-amd64` binaries for all CLIs.

### Checklist

### DIFF
--- a/pkg/gen/input/makefile/internal/file/windows-code-signing.sh.template
+++ b/pkg/gen/input/makefile/internal/file/windows-code-signing.sh.template
@@ -1,0 +1,53 @@
+#!/bin/sh
+
+{{ .Header }}
+
+APPLICATION=$1
+VERSION=$2
+
+SIGNCODE_UTIL=quay.io/giantswarm/signcode-util:1.1.1
+
+echo "APPLICATION=${APPLICATION}"
+echo "VERSION=${VERSION}"
+echo "PWD=${PWD}"
+
+if [ "${CODE_SIGNING_CERT_BUNDLE_PASSWORD}" = "" ]; then
+	echo "Variable CODE_SIGNING_CERT_BUNDLE_PASSWORD not set."
+	exit 1
+fi;
+
+if [ "${CODE_SIGNING_CERT_BUNDLE_BASE64}" = "" ]; then
+	echo "Variable CODE_SIGNING_CERT_BUNDLE_BASE64 not set."
+	exit 1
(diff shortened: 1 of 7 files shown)


marians:
Good call for testing. I found out that because the windows file is a ZIP and not .tar.gz, it can't be found.

marians:
The latest commit represents what has been tested with success in https://github.com/giantswarm/kubectl-gs/runs/6027207195?check_suite_focus=true

anvddriesch (approved):

marians:
Looks like code signing doesn't work as expected yet.

anvddriesch (approved):
Looks good! Let's run it in `kubectl-gs` first to see if it's working as expected and then merge this.

marians:
In kubectl-gs we now have moved the shell code from the Makefile into a [separate script](https://github.com/giantswarm/kubectl-gs/blob/main/.github/windows-code-signing.sh) in the `.github` folder. How can we use devctl to create/maintain that file?

ubergesundheit:
Here is an example PR which shows how you can add files to devctl https://github.com/giantswarm/devctl/pull/379

pkg/gen/input/makefile/internal/file/windows-code-signing.sh.template
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+{{ .Header }}
+
```
ubergesundheit, line 4: Unsure if this script would benefit from `set -e` (http://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#The-Set-Builtin) to make the script exit immediately on any error.

Unsure if this exist in plain /bin/sh

ubergesundheit (approved):
LGTM if tested

marians:
For testing, I am executing it like this:

```
go run main.go gen makefile -f cli -l go
```

however I don't get a `windows-code-signing.sh` file as a result.

ubergesundheit:
Check pkg/gen/gen.go line 52 [link](https://github.com/giantswarm/devctl/blob/master/pkg/gen/gen.go#L52=)

The function returns false and nothing gets done.

So either you change the file name to start with `"zz_generated."` or you add an exception in the function `isRegenerable` (same file)

marians:
Thank you @ubergesundheit ! As always, that was super helpful!

PR #410, opened 2022-04-26
Add permissions field for Input struct and set windows code signing script permissions to 0755

uvegla:
### Context
The generated script for `.github/zz_generated.windows-code-signing.sh` did not have proper quoting and the executable flags were missing from the file.

### Checklist

### DIFF
--- a/pkg/gen/gen.go
+++ b/pkg/gen/gen.go
@@ -2,6 +2,7 @@ package gen
 
 import (
 	"context"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


pkg/gen/input/types.go
```diff
@@ -1,12 +1,18 @@
...
+)
+
 type Input struct {
 	// If delete is true, the file will be deleted if it exists. Allows
 	// for files to be moved/renamed.
 	Delete bool
 	// Path is the absolute path of the file to be generated from this
 	// Input.
 	Path string
+	// Permissions to generate the file with
```
MarcelMue, line 14:
```suggestion
	// Permissions to generate the file with.
```

pkg/gen/gen.go
```diff
@@ -53,7 +54,12 @@ func execute(ctx context.Context, file input.Input) error {
 		return nil
 	}
 
-	w, err := os.OpenFile(file.Path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	var permissions fs.FileMode = 0644
+	if file.Permissions != 0 {
+		permissions = file.Permissions
```
MarcelMue, line 59: I think just setting `file.Permissions = 0644` here would also just work - maybe missing I am missing something

MarcelMue (approved):
Fine with me :)

PR #437, opened 2022-06-27
shell code review

hervenicol:
### Checklist

- quotes for security and reliability
- comments when I needed them to understand
- simplified some bash commands

### DIFF
--- a/pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
@@ -52,8 +52,10 @@ jobs:
           echo "::set-output name=branch::${head}"
 
           head="${head#refs/heads/}" # Strip "refs/heads/" prefix.
+          # If we have at least 2 '#' in ref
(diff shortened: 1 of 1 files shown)


pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
```diff
@@ -52,8 +52,10 @@ jobs:
           echo "::set-output name=branch::${head}"
 
           head="${head#refs/heads/}" # Strip "refs/heads/" prefix.
+          # If we have at least 2 '#' in ref
           if [[ $(echo "$head" | grep -o '#' | wc -l) -gt 1 ]]; then
-            base="$(echo $head | cut -d '#' -f 1)"
```
hervenicol, line 56: As we previously used bash var substitution syntax 2 lines before, I thought it made sense to use it here also.

ubergesundheit:
Thank you for taking a look! Really nice

To test, you can either build a binary from your branch or `go run main.go gen workflows -flavour app` in a repository like https://github.com/ubergesundheit/test-gh-workflows then push :)

I would suggest adding something to the changelog :+1:

Gacko (approved):

PR #438, opened 2022-07-08
Rework helm schema check to make it smarter

glitchcrab:
This PR:

- reworks the helm schema check action to try and cut down on false positives.

#### Rationale

The current version of the schema check is dumb (I know, I wrote it), because it just checks to see if the values file has been altered in the PR and fails if the schema file hasn't been updated. There are many reasons to change a values file (e,g, comments) which mean the file is altered but there is no corresponding change in the schema file. Consequently, this CI check creates a lot of noise and often gets ignored.

This PR makes it much smarter. Instead of just checking if the file is in the PR, it also checks to see if the yaml structure has been changed by comparing it to the values file in the main/master branch. This is achieved with some jq magic which outputs a list of all keys in the file (including nested keys), like so:

```
[
  "name",
  "namespace",
  "image",
  "image.repository",
  "userID",
  "groupID",
  "Installation",
  "Installation.V1",
  "Installation.V1.Registry",
  "Installation.V1.Registry.Domain",
  "port",
  "protocol"
]
```

It then calculates the sha1 hash of the keys and compares it to the hash of the values file in the main/master branch. If the hashes differ then it makes sure the schema file is included in the PR and fails if it isn't.

This isn't perfect, but it's vastly better than the current check.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Align git author identities throughout PR automation workflows
+- Completely rework check_values_schema action to cut down on noise
 
 ## [5.7.0] - 2022-06-23
 
--- a/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
@@ -13,31 +13,66 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          VALUES_FILE_CHANGED="false"
-          SCHEMA_FILE_CHANGED="false"
-
           echo "Comparing ${GITHUB_BASE_REF}...${GITHUB_HEAD_REF}"
 
-          changed_files=$(gh api "repos/{owner}/{repo}/compare/${GITHUB_BASE_REF}...${GITHUB_HEAD_REF}" --jq ".files[] | .filename")
-
+          # check if repo contains a schema file
           if grep -q "values.schema.json" <<< $(git ls-tree -r --name-only ${GITHUB_SHA}); then
 
-            if grep -q "values.yaml" <<< "${changed_files}" ; then
-              VALUES_FILE_CHANGED="true"
+            # get a list of files changed in the PR
+            CHANGED_FILES=$(gh api repos/{owner}/{repo}/compare/${GITHUB_BASE_REF}...${GITHUB_HEAD_REF} \
+              --jq ".files[].filename")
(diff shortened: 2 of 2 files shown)


glitchcrab:
goddamnit nancy :angry:

ubergesundheit:
Thanks for tackling this again! I also tried my luck some months ago, but I think we have to be much smarter than whats implemented here. 

Looks like this will make the action fail if the structure (keys and type of values) of the values.yaml has not been changed (Whitespace or comment changes).

I tried to do this in branch https://github.com/giantswarm/devctl/blob/extend-schema-validation/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template#L53= but its not finished nor working well.

Some other things I notices, (or am wondering about)

- Should we compare against the main/master branch or better the commit the PR was created from?
- One improvement could be to to sort the keys [here](https://github.com/giantswarm/devctl/blob/8c174a8a5d71af14b11f4e32b61f49cd214350e2/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template#L41=) ?
- Does your implementation work for deeply nested keys?
- Does it check the actual types?

glitchcrab:
> Thanks for tackling this again! I also tried my luck some months ago, but I think we have to be much smarter than whats implemented here.
> 
> Looks like this will make the action fail if the structure (keys and type of values) of the values.yaml has not been changed (Whitespace or comment changes).

No, that's exactly what this reworking is meant to avoid. If you check the code snippet I posted above, you can see that it only cares about the keys when comparing the files. If the key structure changes then this will be flagged up by the check, but if you changes don't affect the actual yaml structure then the files will be identical (as far as the check is concerned). You can add as much whitespace and commenting as you like and this check won't care.
 
> I tried to do this in branch https://github.com/giantswarm/devctl/blob/extend-schema-validation/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template#L53= but its not finished nor working well.
> 
> Some other things I notices, (or am wondering about)
> 
>     * Should we compare against the main/master branch or better the commit the PR was created from?

I would be inclined to compare against the default branch because this is our usual workflow - branch from default and then merge in. 

>     * One improvement could be to to sort the keys [here](https://github.com/giantswarm/devctl/blob/8c174a8a5d71af14b11f4e32b61f49cd214350e2/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template#L41=) ?

this is a good idea - I'll change this

>     * Does your implementation work for deeply nested keys?

Yes - check the code snippet in the original PR ;)
 
>     * Does it check the actual types?

No, and I don't think it should. That's the job of `helm lint/template` etc.

ubergesundheit (approved):
After sleeping on this a few days I think you're doing gods work :+1:

PR #439, opened 2022-07-12
Add nancy command to go makefile

uvegla:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `nancy` command to Go makefile for a convenient method to run the Nancy checks the same way as they are done on the CI
+
 ### Changed
 
 - Align git author identities throughout PR automation workflows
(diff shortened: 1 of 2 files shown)


Makefile.gen.go.mk
```diff
@@ -119,6 +119,11 @@ lint: ## Runs golangci-lint.
 	@echo "====> $@"
 	golangci-lint run -E gosec -E goconst --timeout=15m ./...
 
+.PHONY: nancy
```
MarcelMue, line 122: Shouldn't this be done in the template file instead? https://github.com/giantswarm/devctl/blob/master/pkg/gen/input/makefile/internal/file/Makefile.gen.go.mk.template

uvegla: It is done there as well: https://github.com/giantswarm/devctl/pull/439/files#diff-ce91819b452e89cc8bc32e20acd6906351e52a25f03a34392886d23c68a7efc1

Eventually it would have generated it for itself via `github` repository.

uvegla: I kind of added here as well to test the change 😅

MarcelMue (approved):
I missed it the first time around - looks good :)

PR #446, opened 2022-07-27
Bump go module also when releasing a major version with a suffix

fiunchinho:
When creating a new release with a suffix like `1.2.3-alpha1`, the automation did not identify this as a major release, so it didn't bump the go module version.

I changed the check to cover this case.

### Checklist

### DIFF
--- a/pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
@@ -92,7 +92,7 @@ jobs:
             version_major=$(echo "${version}" | cut -d "." -f 1)
             version_minor=$(echo "${version}" | cut -d "." -f 2)
             version_patch=$(echo "${version}" | cut -d "." -f 3)
-            if [[ $version_minor = 0 && $version_patch = 0 ]]; then
+            if [[ $version_minor = 0 && $version_patch =~ ^0.* ]]; then
               echo "::set-output name=is_major::true"
             fi
           fi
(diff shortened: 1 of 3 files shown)


pkg/gen/input/workflows/internal/file/create_release_pr.yaml.template
```diff
@@ -92,7 +92,7 @@ jobs:
             version_major=$(echo "${version}" | cut -d "." -f 1)
             version_minor=$(echo "${version}" | cut -d "." -f 2)
             version_patch=$(echo "${version}" | cut -d "." -f 3)
-            if [[ $version_minor = 0 && $version_patch = 0 ]]; then
+            if [[ $version_minor = 0 && $version_patch =~ ^0.* ]]; then
```
whites11, line 95: technically this will match also `X.0.01`
Is this acceptable?

whites11: Maybe we could use `^0-?.*`.
Or maybe can we simply compare the major version with the currently released major version? Do we have that info?

fiunchinho: > technically this will match also `X.0.01` Is this acceptable?


I don't think that's a valid semver version. According to [the specification](https://semver.org/#spec-item-2)


> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes

whites11: fair enough, even though I think we don't validate that in any way

MarcelMue (approved):
We talked in releng sync, adding a comment and then this can be merged

PR #464, opened 2022-11-04
Simplify the schema check action for helm + check the do the actual check

jkremser:
I like the idea of these checks so I borrowed it for my own repo couple of weeks ago: https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/helm_check-values-schema.yaml

However, I changed it in a way to actutaly perform the validation against the schema, because the previous logic just assumes it passes.

At the same time, this way it's much more simpler, leaving a lot of checks (what files were changed and what not) for the action itself (the paths glob).

Also not using the `gh` cli, but using pure `git` since the repo has been cloned&fetched anyway => saving a call to gh api

Signed-off-by: Jiri Kremser <jiri.kremser@gmail.com>

### Checklist

### DIFF
--- a/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
@@ -1,81 +1,30 @@
 {{{{ .Header }}}}
 name: 'Check if values schema file has been updated'
 on: pull_request
+    branches:
+      - master
+      - main
+    paths:
+      - 'helm/**/values.yaml'
+      - 'helm/**/values.schema.json'
 
 jobs:
   check:
     name: 'Check values.yaml and its schema in PR'
     runs-on: ubuntu-latest
     steps:
-      - name: 'Checkout'
+      - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


kubasobon (approved):

glitchcrab:
seems reasonable to me :+1:

glitchcrab (approved):

pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
```diff
@@ -1,81 +1,30 @@
...
-            fi
-
-          else
-
-            # if grep returns negative then there isn't a values.schema.json to check
-            echo "No values.schema.json file found in branch '${GITHUB_BASE_REF}', nothing to check"
-            exit 0
-
-          fi
+      - name: 'Check if values.json is a valid instance of values.schema.json'
```
marians, line 21: I think this should read

`- name: 'Check if values.yaml is a valid instance of values.schema.json'`

marians: I'll make a PR

PR #471, opened 2022-12-01
Fix job step name

marians:
Adjusting the file name in a step annotation

### Checklist


jkremser (approved):
right, that's better. Internally it does `yq -o=json eval ${VALUES_FILE} > /tmp/values.json` and then does the validation where both files are in json format, but it's an impl. detail

PR #478, opened 2022-12-02
Switch schema validation tool to `yajsv`

marians:
This PR switches the schema validation to https://github.com/neilpa/yajsv

The [previous one](https://github.com/johnnoone/json-spec) had several problems:

- Did not work with the $schema dialect that `helm schema-gen` applies (`"$schema": "http://json-schema.org/schema#"`)
- Did not work with recent JSON Schema dialects, only supported up to draft 4.
- Did not dereference external schema

`yajsv` instead works with all JSON schema dialects. It also directly consumes YAML payloads.

### Checklist

### DIFF
--- a/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
@@ -1,5 +1,5 @@
 {{{{ .Header }}}}
-name: 'Check if values and/or schema file has been updated'
+name: 'Values and schema'
 on:
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


marians:
This is tested in https://github.com/giantswarm/dex-app/pull/234

pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
```diff
@@ -1,5 +1,5 @@
 {{{{ .Header }}}}
-name: 'Check if values schema file has been updated'
+name: 'Values and schema'
```
marians, line 2: Opted for a short title here, as it is displayed in one lone together with the check name:

Values and schema / validate values.yaml against values.schema.json

mogottsch (approved):

PR #483, opened 2022-12-08
create_release_pr: Handle First Release

lyind:
## Problem

The `create_releae_pr` workflows fail in new projects without any pre-existing release, if automatic release numbering is attempted.

Thus it is not possible to simply create a first release `v0.0.1`, `v0.1.0` or `v1.0.0` using the documented `release#major#minor#patch` branch name interface.

This is how the workflow without this PR fails for `release#patch` if no previous release exists:
grafik

## Solution

Handle the `gh api repos/$1/releases/latest` failing and assume version `0.0.0` if the error message contains `Not Found`.

See this screenshot for illustration (successful `release#patch` with this PR applied):
grafik


lyind:
Was discovered during creation of https://github.com/giantswarm/personio-go for giantswarm/giantswarm#24049

ced0ps (approved):
LGTMIT

PR #512, opened 2023-02-01
Improve create_release_pr

erkanerol:
We noticed that the chart version for https://github.com/giantswarm/cluster-api-provider-cloud-director-app is not updated by CI automatically. Then I checked the CI logs and found this.

Screenshot 2023-02-01 at 13 36 36

Basically, we have different cases regarding repo name vs app name
1. AppName = RepoName
2. AppName = RepoName+"-app"
3. AppName+"-app" = RepoName

1 and 2 are already handled. 3 is not handled in the current form. This PR handles it too. 

Also note that it is not only a problem in https://github.com/giantswarm/cluster-api-provider-cloud-director-app. I checked other providers and they are the same too. Example https://github.com/giantswarm/cluster-api-provider-aws-app

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


erkanerol:
NEVERMIND

PR #528, opened 2023-03-07
Add workflows and makefile for cluster app

mogottsch:
This PR adds a new flavour for cluster apps together with the related workflows (GitHub actions) and a Makefile.
The GitHub actions and the makefile revolve around the `values.schema.json` and `values.yaml`.

The Makefile includes commands to normalize and verify the `values.schema.json` file with [schemalint](https://github.com/giantswarm/schemalint), as well as, a command to generate `values.yaml` from `values.schema.json` with [helm-values-gen](https://github.com/giantswarm/helm-values-gen).

The GitHub actions check (on PR) that `values.schema.json` is verified and that `values.yaml` is generated from `values.schema.json`.

## Context
Towards https://github.com/giantswarm/roadmap/issues/2093

The idea is to add `cluster-app` as a flavor to all cluster app repositories (`cluster-aws`, `cluster-azure`, ...).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Add new flavour with workflows and makefile for cluster app
+- Add new flavour with workflows and makefile for cluster apps.
 
 ## [5.19.0] - 2023-02-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add new flavour with workflows and makefile for cluster app
+
 ## [5.19.0] - 2023-02-21
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
 - Add new flavour with workflows and makefile for cluster apps.
+
+### Changed
+
+- repo setup: better select checks required for PR merge
 
 ## [5.19.0] - 2023-02-21
 
--- a/cmd/gen/makefile/runner.go
+++ b/cmd/gen/makefile/runner.go
@@ -64,6 +64,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			inputs = append(inputs, in.MakefileGenKubernetesAPI()...)
 		}
 
+		if r.flag.Flavours.Contains(gen.FlavourClusterApp) {
+			inputs = append(inputs, in.MakefileGenClusterApp())
+		}
+
 		if r.flag.Language == gen.LanguageGo {
 			inputs = append(inputs, in.MakefileGenGo()...)
 		}
--- a/cmd/gen/workflows/runner.go
+++ b/cmd/gen/workflows/runner.go
@@ -77,7 +77,7 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 	}
 
 	if r.flag.Flavours.Contains(gen.FlavourClusterApp) {
-		inputs = append(inputs, workflowsInput.JSONSchemaValidation())
+		inputs = append(inputs, workflowsInput.ClusterAppSchemaValidation())
 	}
 
 	err = gen.Execute(
--- a/cmd/gen/workflows/runner.go
+++ b/cmd/gen/workflows/runner.go
@@ -76,6 +76,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		inputs = append(inputs, workflowsInput.AddCustomerBoardAutomation())
 	}
 
+	if r.flag.Flavours.Contains(gen.FlavourClusterApp) {
+		inputs = append(inputs, workflowsInput.JSONSchemaValidation())
+	}
+
 	err = gen.Execute(
 		ctx,
 		inputs...,
--- a/pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
+++ b/pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
@@ -6,16 +6,16 @@ VALUES_SCHEMA=$(shell find ./helm -maxdepth 2 -name values.schema.json)
 
 .PHONY: normalize-schema
 normalize-schema:
-	go install github.com/giantswarm/schemalint@latest
+	go install github.com/giantswarm/schemalint@v1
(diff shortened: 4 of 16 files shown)


pkg/gen/input/workflows/internal/file/json_schema_validation.yaml.template
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "^1.19"
+
+      - name: Install schemalint
+        run: |
+          go install github.com/giantswarm/schemalint@latest
```
ubergesundheit, line 29: IMO we shouldn't install `latest` here to not break peoples builds in case the latest version of schemalint introduces some breaking changes.

Also other workflows use https://github.com/giantswarm/install-binary-action to install pre-built binaries in GitHub actions.

pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
```diff
@@ -0,0 +1,21 @@
+{{ .Header }}
+
+
+VALUES=$(shell find ./helm -maxdepth 2 -name values.yaml)
+VALUES_SCHEMA=$(shell find ./helm -maxdepth 2 -name values.schema.json)
+
+.PHONY: normalize-schema
+normalize-schema:
+	go install github.com/giantswarm/schemalint@latest
```
ubergesundheit, line 9: Also here. I think using `latest` here and later in this file is not a good idea

cmd/gen/workflows/runner.go
```diff
@@ -76,6 +76,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		inputs = append(inputs, workflowsInput.AddCustomerBoardAutomation())
 	}
 
+	if r.flag.Flavours.Contains(gen.FlavourClusterApp) {
```
ubergesundheit, line 79: needs documentation in `gen workflows` help

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add new flavour with workflows and makefile for cluster app
```
ubergesundheit, line 10: What is a cluster app? When would devs use this makefile or workflow

pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
ubergesundheit: Other makefiles use `docker run` instead of executing `go install`. Debatable if running a container is more portable/better/anything.

kopiczko:
What is the difference between `cluster-app` and `app`? Shouldn't we add schema validation to all apps?

pkg/gen/input/workflows/internal/file/json_schema_validation.yaml.template
mogottsch: Nice idea, I'm looking into it :+1:

mogottsch:
> What is the difference between `cluster-app` and `app`? Shouldn't we add schema validation to all apps?

Currently, we impose some requirements on the `values.schema.json` of cluster apps (`cluster-azure`, `cluster-aws`, ...). You can see them [here in this RFC](https://github.com/giantswarm/rfc/pull/55). These requirements are checked by schemalint.
They are generally very strict and it takes some effort to migrate the `values.schema.json` to conform with these requirements.
Also, it includes some requirements, that are only meant for cluster apps.

In the future, I think it would definitely be nice to use schemalint and helm-values-gen for all apps, but that would take a lot of effort.

Also, the overall goal that drives this endeavor is the consistent creation of clusters, through the different UIs. We haven't really talked about other apps, so far.

pkg/gen/input/workflows/internal/file/json_schema_validation.yaml.template
mogottsch: I created a composite action in the schemalint repository and now use it here.
I've also requested your review on [the PR](https://github.com/giantswarm/schemalint/pull/57) that adds the action to the schemalint repository. I would appreciate if you could take a look at it!

CHANGELOG.md
mogottsch: With a cluster apps, I mean repositories like `cluster-aws`, `cluster-azure`, etc.

A similar version of the Makefile and workflows of this PR is already in use in [cluster-azure](https://github.com/giantswarm/cluster-azure).
You can take a look here:
[Makefile](https://github.com/giantswarm/cluster-azure/blob/main/Makefile.schema.mk)
[workflow](https://github.com/giantswarm/cluster-azure/blob/main/.github/workflows/json_schema.yaml)

Also for a explanation why we differentiate between a cluster-app and a regular app take a look at [this comment](https://github.com/giantswarm/devctl/pull/528#issuecomment-1459876880).

ubergesundheit: thanks for the explanation. I was not aware of the term.

mogottsch:
My current plan for the version management of `schemalint` and `helm-values-gen` in the makefiles and the workflows is the following.
- Schemalint and helm-values-gen both will use [floating tags](https://github.com/giantswarm/floating-tags-action).
- Devctl will point to the floating major versions, e.g. `v1`.
- When a new major release is created, we have to manually update the version of the respective repo in devctl.

That way minor and patch releases of helm-values-gen and devctl will be rolled automatically, so that we don't flood the repositories using these makefiles and workflows with PRs. Major releases won't be rolled automatically but only after updating devctl and the respective "Align files" PRs. This allows us to keep control over rolling out large chunks of breaking changes.

pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
mogottsch: I replaced `latest` with a major floating tag.
Concerning `docker run`: I will leave this as-is for now. If someone complains, it will still be possible to adapt.

cmd/gen/makefile/runner.go
```diff
@@ -64,6 +64,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			inputs = append(inputs, in.MakefileGenKubernetesAPI()...)
 		}
 
+		if r.flag.Flavours.Contains(gen.FlavourClusterApp) {
```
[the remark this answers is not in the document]
mogottsch: done

cmd/gen/workflows/runner.go
mogottsch: done

marians (approved):

mogottsch:
Everything is set now for this PR to be merged.
I also tested the the changes by running `devctl gen makefile` and `devctl gen workflows` on a [branch](https://github.com/giantswarm/cluster-azure/tree/test-new-json-schema-workflow) of cluster-azure.

PR #535, opened 2023-03-23
repo setup: filter aliyun checks out of required checks for PR merge

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/24617

Aliyun jobs are and not systematically triggered.
So we should keep them as optional when updating list of required checks for PRs. 

I tried to do it so that we could add more exclusions later.

### Checklist

### DIFF
--- a/cmd/repo/setup/flag.go
+++ b/cmd/repo/setup/flag.go
@@ -28,7 +28,8 @@ type flag struct {
 	Permissions map[string]string
 
 	// Branch protection
-	Checks []string
+	Checks       []string
+	ChecksFilter string
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
@@ -56,6 +57,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// Branch protection
 	cmd.PersistentFlags().StringSliceVar(&f.Checks, "checks", nil, "Check context names for branch protection. Default will add all auto-detected checks, this can be disabled by passing an empty string.")
+	cmd.PersistentFlags().StringVar(&f.ChecksFilter, "checks-filter", "aliyun", "Provide a regex to filter checks. Checks matching the regex will be ignored.")
 }
 
 func (f *flag) Validate() error {
--- a/cmd/repo/setup/flag.go
+++ b/cmd/repo/setup/flag.go
@@ -56,7 +56,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringToStringVar(&f.Permissions, "permissions", map[string]string{"Employees": "admin", "bots": "push"}, "Grant access to this repository using github_team_name=permission format. Multiple values can be provided as a comma separated list or using this flag multiple times. Permission can be one of: pull, push, admin, maintain, triage.")
 
 	// Branch protection
-	cmd.PersistentFlags().StringSliceVar(&f.Checks, "checks", nil, "Check context names for branch protection. Default will add all auto-detected checks, this can be disabled by passing an empty string.")
+	cmd.PersistentFlags().StringSliceVar(&f.Checks, "checks", nil, "Check context names for branch protection. Default will add all auto-detected checks, this can be disabled by passing an empty string. Overrides \"--checks-filter\"")
 	cmd.PersistentFlags().StringVar(&f.ChecksFilter, "checks-filter", "aliyun", "Provide a regex to filter checks. Checks matching the regex will be ignored. Empty string disables filter (all checks are accepted).")
 }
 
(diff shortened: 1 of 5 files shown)


pkg/githubclient/client_repository.go
```diff
@@ -247,15 +248,34 @@ func (c *Client) getGithubChecks(ctx context.Context, repository *github.Reposit
...
+			if filterCheck(status.GetContext()) {
+				checks = append(checks, status.GetContext())
+			}
 		}
 	}
 
 	c.logger.Debugf("found %d commit statuses for ref %q:\n%v", len(checks), ref, checks)
+	for id, check := range checks {
+		c.logger.Debugf(" - checks[%d] = %q", id, check)
+	}
```
TheoBrigitte, line 260: Now we are printing `checks` twice.

Let's make a choice, either keep it in the first (257) line, or keep the loop but not both.
I am fine either way.

pkg/githubclient/client_repository.go
```diff
@@ -247,15 +248,34 @@ func (c *Client) getGithubChecks(ctx context.Context, repository *github.Reposit
...
 
 	return checks, nil
 }
 
+// Returns false if check should be ignored
+func filterCheck(checkName string) bool {
+	// List keywords for checks that should be ignored
+	var ignoredCheckKeywords = []string{"aliyun"}
```
TheoBrigitte, line 268: I am not a fan of hiding values deep down in code.

Please move this into a command flag, we can keep `aliyun` as default value.

pkg/githubclient/client_repository.go
```diff
@@ -247,15 +248,34 @@ func (c *Client) getGithubChecks(ctx context.Context, repository *github.Reposit
...
 
 	c.logger.Debugf("found %d commit statuses for ref %q:\n%v", len(checks), ref, checks)
+	for id, check := range checks {
+		c.logger.Debugf(" - checks[%d] = %q", id, check)
+	}
 
 	return checks, nil
 }
 
+// Returns false if check should be ignored
```
TheoBrigitte, line 265: Correct Golang comment syntax for function is `// functionName comment`

```suggestion
// filterCheck returns false if check should be ignored
```

pkg/githubclient/client_repository.go
```diff
@@ -247,15 +248,34 @@ func (c *Client) getGithubChecks(ctx context.Context, repository *github.Reposit
...
+func filterCheck(checkName string) bool {
+	// List keywords for checks that should be ignored
+	var ignoredCheckKeywords = []string{"aliyun"}
+
+	for _, ignoredCheckKeyword := range ignoredCheckKeywords {
+		matched, _ := regexp.MatchString(ignoredCheckKeyword, checkName)
```
TheoBrigitte, line 271: Let's not ignore the returned error, this is bad practice.
We can also compile the regex beforehand (I'm thinking when we get the flag value), and just use `MatchString(string) bool` method here.

TheoBrigitte (changes requested):
Nice improvement, but let's make it configurable instead of hardcoding it.

cmd/repo/setup/flag.go
```diff
@@ -56,6 +57,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// Branch protection
 	cmd.PersistentFlags().StringSliceVar(&f.Checks, "checks", nil, "Check context names for branch protection. Default will add all auto-detected checks, this can be disabled by passing an empty string.")
+	cmd.PersistentFlags().StringVar(&f.ChecksFilter, "checks-filter", "aliyun", "Provide a regex to filter checks. Checks matching the regex will be ignored.")
```
TheoBrigitte, line 60: This only works when `--checks` is not set. We need to mention this here.

cmd/repo/setup/runner.go
```diff
@@ -67,6 +68,14 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
...
 	}
 
+	ChecksFilterRegexp, err := regexp.Compile(r.flag.ChecksFilter)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	if r.flag.ChecksFilter == "" {
+		ChecksFilterRegexp = nil
+	}
+
```
TheoBrigitte, line 78: I would prefer this, so we do not even bother compiling the regexp when we ignore the filter.

```suggestion
	var ChecksFilterRegexp *regexp.Regexp
	if r.flag.ChecksFilter != "" {
		ChecksFilterRegexp, err := regexp.Compile(r.flag.ChecksFilter)
		if err != nil {
			return microerror.Mask(err)
		}
	}
```

TheoBrigitte (approved):
double approval

PR #556, opened 2023-04-12
Check values schema on push, not only for PRs

AndiDog:
https://github.com/giantswarm/roadmap/issues/2144

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `--disable-branch-protetion` flag to `devctl repo setup`, to allow disabling github branch protection.
 
+### Changed
+
+- Check values schema on push, not only for PRs
+
 ## [5.22.0] - 2023-04-13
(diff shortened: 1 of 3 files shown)


fiunchinho (approved):

AndiDog:
Need to change this again because `GITHUB_BASE_REF` is only available in PRs, not on push. Tested this with `cluster-aws`.

Can @giantswarm/team-rainbow take a look?

mogottsch (commented):
The workflow file you modified only checks whether `values.yaml` is a valid payload of the JSON schema defined in `values.schema.json`. 
The error message you posted in slack (`values.yaml is out of sync with values.schema.json...`), is from [this workflow](https://github.com/giantswarm/devctl/blob/main/pkg/gen/input/workflows/internal/file/cluster_app_schema_validation.yaml.template).
I think it makes sense to run that on push as well.

AndiDog:
@mogottsch great idea – I implemented and tested that successfully

marians:
FYI This action does not check whether `values.schema.json` exists

PR #583, opened 2023-06-01
Add make target 'generate-docs' for cluster apps

marians:
Towards https://github.com/giantswarm/roadmap/issues/1733

This PR adds the make target `make generate-docs` to the Makefile for repos of flavor `cluster-app`. This facilitates updating the docs for cluster app maintainers.

### Checklist

### DIFF
--- a/pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
+++ b/pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
@@ -4,6 +4,7 @@
 
 VALUES=$(shell find ./helm -maxdepth 2 -name values.yaml)
 VALUES_SCHEMA=$(shell find ./helm -maxdepth 2 -name values.schema.json)
+CHART_README=$(shell find ./helm -maxdepth 2 -name README.md)
 
 .PHONY: normalize-schema
 normalize-schema: ## Normalize the JSON schema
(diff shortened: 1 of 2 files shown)


pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
```diff
@@ -15,6 +16,11 @@ validate-schema: ## Validate the JSON schema
 	go install github.com/giantswarm/schemalint/v2@v2
 	schemalint verify $(VALUES_SCHEMA) --rule-set=cluster-app
 
+.PHONY: generate-docs
+generate-docs: ## Generate values documentation from schema
+	go install github.com/giantswarm/schemadocs@latest
```
mogottsch, line 21: Is schemadocs already used in CI? If so `latest` could be problematic.

For `schemalint` and `helm-values-gen` we don't use `latest` but always point to the latest major version tag, which is generated by this action: https://github.com/giantswarm/floating-tags-action

This is an additional effort, but it has the advantage that we can control when major changes of the tools are rolled out.

Whether or not this is useful, depends on whether `schemadocs` is used in CI, to check the generated README.
If it is used there could be a version mismatch between the one in the CI and the one the developer uses locally.

marians: So far, as far as I know, schemadocs isn't used in CI yet.

schemadocs is at v0.0.1. Which is the right major version then to use here?

mogottsch: If it's not used in CI, I think it's fine to keep `latest`.

To use the floating major version tags (v0, v1, v2, etc...), you would have to install the action for this repo.
To do that, you have to add `enableFloatingMajorTags: true` in the `team-honeybadger.yaml` in the github repository.
Then we could use `@v0` here instead of `@latest`. But that also means that you have to manually update the version here, every time a new major comes out.

mogottsch (approved):

pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template
marians: Thanks!

PR #595, opened 2023-07-05
Changelog: Remove `nginx-ingress-controller`.

Gacko:
### Checklist


ubergesundheit:
Maybe this can be removed altogether? Looks like this was required when ingress-nginx was a default app.

Gacko:
Hm, ok. Checked it. Really seems to be involved for forging releases only. Gonna remove it here.

Gacko:
Removed it. Thanks for pointing this out! :)

MarcelMue (approved):

PR #621, opened 2023-09-14
Make renovate ignore the module `github.com/imdario/mergo`

marians:
The module got renamed, but Renovate tries to update it to v1.0.0 using the old name, which fails.

As the module is in use in 166 repos, we exclude this globally to prevent extra handling of this.

### Checklist


marians:
@ljakimczuk Do you think we should add more dependencies to be ignored?

ljakimczuk (approved):

PR #625, opened 2023-09-15
Update action/checkout to v4

mcharriere:
- update `action/checkout` to v4 in generated files
- update changelog

### Checklist


lyind (commented):
LGTM

Maybe the dependencies could be declared in such a way that `dependabot` sees and updates them?

Then we would decrease the likelyhood of `dependabot` and `devctl` fighting each other.

Would it be enough to store the workflow templates somewhere where `dependabot` can see them or maybe create some config file for `dependabot`?

marians (approved):

Gacko (approved):

PR #660, opened 2023-10-17
Use our own Gitleaks action

Strigix:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


kopiczko:
Is there an issue or explanation linked to that?

piontec (approved):
LGTM, if we can't use upstream anymore :)

Strigix:
@kopiczko 

It's because of this: https://github.com/giantswarm/giantswarm/issues/27966

basically, gitleaks-action went from MIT license to a paid one (with wicked prices).
So we were still using the older one, but it had a deprecated feature (set-output)
I copied the last one under MIT license (v1.6) and removed the deprecated command.

PR #750, opened 2023-12-14
Add OSSF Scorecard workflow to devctl gen workflows

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/24297

### Checklist

### DIFF
--- a/pkg/gen/input/workflows/internal/file/run_ossf_scorecard.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/run_ossf_scorecard.yaml.template
@@ -1,0 +1,73 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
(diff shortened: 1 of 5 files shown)


stone-z:
Tested in https://github.com/giantswarm/github/pull/1300

stone-z:
Auto-merging to not miss a potential devctl release while I'm away

pkg/gen/input/workflows/internal/file/run_ossf_scorecard.yaml.template
```diff
@@ -0,0 +1,75 @@
...
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '15 15 15 * *'
+  push:
+    branches: [ "main" ]
```
marians, line 17: Unfortunately, quite some repos still have their default branch named `master`

stone-z: Ah yes I didn't think about that. It wouldn't be great to have to specify the branch manually for each repo either.
When I return I'll see about detecting the default branch

stone-z: I checked quickly and it is possible ot detect the default branch through the GitHub API (`gh repo view --json defaultBranchName`) but it also works to simply specify both branches, as long as there are only pushes to the actual default branch. In other words, the action will complain if the default branch is `main` but there are additional pushes to `master`. I will update this to list both branches

stone-z:
Ready for review now

marians (approved):

PR #766, opened 2024-01-09
Add workflow for linting CHANGELOG.md

marians:
Towards https://github.com/giantswarm/giantswarm/issues/29422

This PR adds a workflow to be added to all flavours of GitHub project repos. The workflow is called `lint-changelog` and checks the `CHANGELOG.md` file if it has been verified in the context of a pull requests.

The workflow is intended to be used as an optional (not required) check for pull requests, since some PRs, e. g. those made by Renovate, won't get changelog updates.

What is checked?

- Check if the changelog is empty
- Check if the changelog can be parsed using [parse-changelog](https://github.com/cseelye/parse-changelog)
- Check if there are unreleased changes mentioned, as should normally be the case for a pull request

### Checklist


pkg/gen/input/workflows/internal/file/lint_changelog.yaml.template
```diff
@@ -0,0 +1,62 @@
...
+        else
+          echo "Found $LENGTH top-level items"
+        fi
+    - name: Check for unreleased changes
+      run: |
+        CHANGES=$(parse-changelog --changelog ./CHANGELOG.md --show-unreleased | tr -d '[:space:]')
+        if [ -z "$CHANGES" ]
+        then
+          echo "::warning file=CHANGELOG.md,line=1,col=1,endColumn=1::Please don't forget to add info about your changes to CHANGELOG.md"
+        fi
```
AndiDog, line 62: Could we check for trailing whitespace, maybe? People sometimes introduce those, and they get trimmed away by reasonably-configured editors and that can be conflicting/annoying on rebase. I'd like to avoid that overall, even if it sounds minor.

marians: So basically we want strict Markdown linting on changelogs?

AndiDog: Markdown allows extra whitespace, and even gives it special meaning (google "markdown two spaces"), but trailing whitespace should be avoided everywhere since some people configure their editors to trim them correctly (and everyone should). So it's not about Markdown linting, but I'd be in favor of running `markdownlint` against CHANGELOG.md if the findings are shown in GitHub PRs.

pipo02mix:
so we close?

marians:
This became obsolete

PR #825, opened 2024-02-17
Rename fix vulnerabilities workflow to remediate vulnerabilities

fhielpos:


pipo02mix:
@fhielpos do you want to go ahead?

fhielpos:
I was having issues with running the docker image in some repos so I think I will shelve this until I come up with a better approach

PR #826, opened 2024-02-19
OSSF Scorecard: Update CodeQL action.

Gacko:
We are currently receiving Dependabot PRs for this in repositories managed by `giantswarm/github`. This PR aligns the managed configuration to the actual one.

### DIFF
--- a/pkg/gen/input/workflows/internal/file/run_ossf_scorecard.yaml.template
+++ b/pkg/gen/input/workflows/internal/file/run_ossf_scorecard.yaml.template
@@ -70,6 +70,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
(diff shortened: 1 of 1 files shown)


marians:
@Gacko In that case, we should configure our presets so that Renovate ignores generated files. Can you please point me to a Renovate PR?

Gacko:
Sure, this one: https://github.com/giantswarm/cluster-autoscaler-app/pull/254#pullrequestreview-1887777606

Gacko:
But this PR here in particular is ok so far?

marians:
> But this PR here in particular is ok so far?

I can't tell, to be honest. That's why I have requested Team Shield to review.

Gacko:
> That was dependabot, not Renovate

Yeah, sorry, my fault. Again. Turns out you were right, I mixed that up because there were so many PRs this morning, some from Renovate but bumping something else, some from Dependabot bumping generated files.

So to get that clear: This dependency in particular got bumped only by Dependabot and we can in fact remove it (Dependabot) from those repositories. Will care for that later.

PR #894, opened 2024-04-03
Change generated file headers to include GitHub link instead of version

ubergesundheit:
This PR changes the header of generated files to contain a GitHub link instead of the devctl version.

Follow up of #875

Example:
```
# DO NOT EDIT. Generated with:
#
#    devctl
#
#    https://github.com/giantswarm/devctl/blob/6a704f7e2a8b0f09e82b5bab88f17971af849711/pkg/gen/input/makefile/internal/file/Makefile.template
#
```

The URL points to the template on GitHub using the latest commit the template was changed.

IMO this helps in multiple ways:

- One can use the blame feature to see when a specific line in the template was changed and why (browsing history of the template at the source)
- In case a template needs to be changed, its easier to find the exact file to change

The changes in this PR can be tested by using the container image like this:
```
docker run --rm -it -v $(pwd):/app -w /app \
  quay.io/giantswarm/devctl:6.23.3-784664fc1434a6494500bc0eff8bee9e8d4bebfa \
  gen workflows -l go -f cli
```

/cc @lyind @marians as you were involved in the last PR

### Checklist


marians:
OK, first of all, it seems kind of cool!

Am I right assuming that the URL like `https://github.com/giantswarm/devctl/blob/6a704f7e2a8b0f09e82b5bab88f17971af849711/pkg/gen/input/makefile/internal/file/Makefile.template` always represents the last commit that modified the template? So there will be no PRs which only update the header?

Second, I am not exactly sure how useful that URL with a git commit is.

ubergesundheit:
Yes, the URL points to the template on GitHub using the latest commit the template was changed.

IMO this helps in multiple ways:

- One can use the blame feature to see when a specific line in the template was changed and why (browsing history of the template at the source)
- In case a template needs to be changed, its easier to find the exact file to change

marians (approved):
OK, fair enough! Let's try it out.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Header of generated files do not contain the devctl version any more
```
lyind, line 12:
```suggestion
- devctl version is no longer added to generated file header
```

lyind (approved):
𝕃𝔾𝕋𝕄

Only minor suggestion in Changelog (spelling)

PR #942, opened 2024-04-25
Fix generated headers: Fetch whole git history for devctl

ubergesundheit:
This PR is a fix and follow up to #894

I suspect missing git history causing wrong git urls generated. As of now the header urls contain the commit of the release. We want the commit which last changed a file.

### Checklist


marians (approved):
Nice!

PR #1265, opened 2025-03-27
Add a command to deploy an app to a GitOps repo

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/32941

### Checklist


cmd/deploy/command.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	c := &cobra.Command{
+		Use:   "deploy",
+		Short: "Deploy an application to a GitOps repository",
+		Long: `Deploy an application to a GitOps repository by:
+		1. Cloning the GitOps repository
+		2. Adding the application using 'kubectl gs gitops add app'
+		3. Creating a branch and committing the changes
+		4. Creating a pull request
```
pipo02mix, line 44:
```suggestion
		4. Creating a pull request
		5. Check the app is finally deployed
```

pkg/appstatus/client.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return c, nil
+}
+
+func (c *Client) WaitForAppDeployment(ctx context.Context, appName, orgNamespace, managementCluster string, timeout time.Duration) error {
+	c.logger.Infof("Waiting for app %s to be deployed in namespace %s", appName, orgNamespace)
+
+	// Login to management cluster
+	loginCmd := exec.Command("tsh", "kube", "login", managementCluster)
```
lyind, line 43: I'm not sure I like tools to play with `kubectl` state like that, but could find a better way after reading `tsh kube login` docs:
```suggestion
	loginCmd := exec.Command("tsh", "kube", "login", managementCluster)
```

pkg/githubclient/client.go
```diff
@@ -2,46 +2,171 @@ package githubclient
...
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
-
 	if config.AccessToken == "" {
 		return nil, microerror.Maskf(invalidConfigError, "%T.AccessToken must not be empty", config)
 	}
 
 	c := &Client{
-		dryRun: config.DryRun,
```
lyind, line 36: What's it with the `dry-run` feature? It doesn't make sense anymore (understandable, as operations are complex and depend on each other)?

lyind (approved):
I don't like that external state of `kubectl` is modified via `tsh kube login ...`, but it seems acceptable.

LGTM 👍

pkg/githubclient/client_repository.go
```diff
@@ -201,12 +197,10 @@ func (c *Client) SetRepositoryBranchProtection(ctx context.Context, repository *
...
 	c.logger.Debugf("branch protection settings\n%s", b)
 
-	if !c.dryRun {
-		underlyingClient := c.getUnderlyingClient(ctx)
-		_, _, err = underlyingClient.Repositories.UpdateBranchProtection(ctx, owner, repo, default_branch, opts)
-		if err != nil {
-			return microerror.Mask(err)
-		}
+	underlyingClient := c.getUnderlyingClient(ctx)
+	_, _, err = underlyingClient.Repositories.UpdateBranchProtection(ctx, owner, repo, default_branch, opts)
```
lyind, line 201: So every repository touched gets branch protection enabled? Is this really a good idea (it feels quite opinionated)?
```suggestion
	_, _, err = underlyingClient.Repositories.UpdateBranchProtection(ctx, owner, repo, default_branch, opts)
```

lyind (approved):

pkg/appstatus/client.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		c.logger.Infof("App %s is not deployed yet, current status: %s", appName, status)
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			continue
+		}
+	}
```
TheoBrigitte, line 90: Simplifying this by using a context with timeout and using a go routine for the app status check

```suggestion
	ctx, cancel := context.WithTimeout(ctx, timeout)
	
	go func(appName, orgNamespace string) {
		// Get app status using kubectl
		kubectlCmd := exec.Command("kubectl", "get", "app", appName,
			"-n", orgNamespace,
			"-o", "jsonpath={.status.release.status}")
		kubectlCmd.Stderr = c.stderr

		output, err := kubectlCmd.Output()
		if err == nil {
			status := string(output)
			if status == "deployed" {
				c.logger.Infof("App %s successfully deployed", appName)
				cancel()
				return
			}
			c.logger.Infof("App %s is not deployed yet, current status: %s", appName, status)
		 }

		 <-ticker.C
	}(appName, orgNamespace)

	
	<-ctx.Done()
	return ctx.Err()
```

pkg/githubclient/client.go
```diff
@@ -2,46 +2,171 @@ package githubclient
...
+			return nil
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			continue
+		}
+	}
```
TheoBrigitte, line 167: This can use the same context timeout logic I suggested earlier.

cmd/repo/setup/ciwebhooks/runner.go
```diff
@@ -52,7 +52,6 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 	}
 
 	c := githubclient.Config{
-		DryRun:      r.flag.DryRun,
```
TheoBrigitte, line 55: Why killing all the dry runs ?

TheoBrigitte (changes requested):

cmd/repo/setup/ciwebhooks/runner.go
pipo02mix: hmm I think it was not intentional 😅

pipo02mix:
> I don't like that external state of kubectl is modified via tsh kube login ..., but it seems acceptable.

yeah I was thinking on generating a specific kubeconfig and run it as arg in kubectl command

lyind:
> yeah I was thinking on generating a specific kubeconfig and run it as arg in kubectl command

lyind:
> yeah I was thinking on generating a specific kubeconfig and run it as arg in kubectl command

This would be perfect. Maybe create a note or a separete issue referencing this one?

pkg/githubclient/client_repository.go
pipo02mix: I did not modify the behavior, just removed the dry run

lyind: OK, then I guess it's intended/approved.

pkg/appstatus/client.go
```diff
@@ -41,51 +41,37 @@ func (c *Client) WaitForAppDeployment(ctx context.Context, appName, orgNamespace
...
 	defer ticker.Stop()
 
-	deadline := time.Now().Add(timeout)
-	for {
-		if time.Now().After(deadline) {
-			return microerror.Maskf(executionFailedError, "timeout waiting for app %q to be deployed", appName)
-		}
+	ctx, cancel := context.WithTimeout(ctx, timeout)
 
+	go func(appName, orgNamespace string) {
```
TheoBrigitte, line 54: My bad I forgot the `for` loop inside this function.

pkg/githubclient/client.go
```diff
@@ -143,28 +143,29 @@ func (c *Client) WaitForPRMerge(ctx context.Context, owner, repo string, prNumbe
 	ticker := time.NewTicker(10 * time.Second)
 	defer ticker.Stop()
 
-	deadline := time.Now().Add(timeout)
-	for {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+
+	go func(owner, repo string, prNumber int) {
```
TheoBrigitte, line 148: You also need a for loop inside this function to keep retrying.

And this will likely change the return for continue directives

pipo02mix: ok I was confused how that would work, but did not have time to test it yet. today I will do

pkg/appstatus/client.go
```diff
@@ -0,0 +1,90 @@
...
+			<-ticker.C
+		}
+	}(appName, orgNamespace)
+
+	<-ctx.Done()
+	err = ctx.Err()
+	if err == context.Canceled {
+		return nil
+	}
+	return ctx.Err()
```
pipo02mix, line 89: @TheoBrigitte not sure how to make this cleaner, suggestions?

pkg/appstatus/client.go
```diff
@@ -50,28 +50,41 @@ func (c *Client) WaitForAppDeployment(ctx context.Context, appName, orgNamespace
...
-		kubectlCmd.Stderr = c.stderr
-
-		output, err := kubectlCmd.Output()
-		if err == nil {
-			status := string(output)
-			if status == "deployed" {
-				c.logger.Infof("App %s successfully deployed", appName)
+		for {
+			if time.Now().After(deadline) {
+				c.logger.Errorf("App %s was not deployed within %v", appName, timeout)
```
TheoBrigitte, line 58: You don't need this as the timeout is already in the context. But we could add a check for ctx.Done to ensure the goroutine is actually stopped when timeout is reached.

Somethine like


```go
	select {
  		case <-ctx.Done():
  			return ctx.Err()
  		case out <- ticker.C:
			// nothing here, let the function run.
  	}
```

pkg/appstatus/client.go
TheoBrigitte: I think that's fine

```suggestion
	<-ctx.Done()

	err = ctx.Err()
	if errors.Is(err, context.Canceled) {
		return nil
	}
	return err
```

cmd/deploy/runner.go
```diff
@@ -33,6 +33,24 @@ func (r *runner) Run(cmd *cobra.Command, args []string) error {
...
+	if _, err := exec.LookPath("kubectl-gs"); err != nil {
+		fmt.Fprintln(os.Stderr, "Error: 'kubectl gs' plugin is not installed or not found in PATH")
+		os.Exit(1) // Exit the program with a non-zero status
+	}
+
+	// Check if 'tsh' is installed
+	if _, err := exec.LookPath("tsh"); err != nil {
+		fmt.Fprintln(os.Stderr, "Error: 'tsh' is not installed or not found in PATH")
+		os.Exit(1) // Exit the program with a non-zero status
+	}
```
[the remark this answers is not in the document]
pipo02mix: hmm I was doubting, ok I remove it :)

TheoBrigitte (approved):
LGTM

pkg/githubclient/client.go
```diff
@@ -2,46 +2,173 @@ package githubclient
...
+
+		select {
+		case <-ticker.C:
+			c.logger.Infof("Waiting for PR #%d to be merged", prNumber)
+		case <-ctx.Done():
+			return microerror.Maskf(invalidConfigError, "PR #%d was not merged within %v", prNumber, timeout)
+		}
+	}
+}
+
```
pipo02mix, line 171: @TheoBrigitte I could not make it work with a `go func`, and I don't know why you would use a coroutine for a command. Like this works fine. WDYT?

PR #1284, opened 2025-04-22
Update ubuntu to a fixed version in CI

pipo02mix:
### Checklist


marians (changes requested):
According to [this](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories) it should be `ubuntu-24.04` (with a period)

pipo02mix:
Good catch, I have tested in the [debug-toolbox](https://github.com/giantswarm/debug-toolbox/tree/main/.github/workflows) manually but then I failed to replace with correct name :facepalm

marians (approved):

PR #1395, opened 2025-07-25
Add LLM agent rule generation

stone-z:
2025 hAIve sprint hackathon

Generates a base set of LLM rules that try to get LLMs to use our central style guide located in `fmt`.

Supports additional generation of language- or flavour-specific rules. It is not the intent to store specific rules here in devctl -- rather, these specific rulesets should be used to direct / target the agent to resources elsewhere.

### Checklist

### DIFF
--- a/cmd/gen/command.go
+++ b/cmd/gen/command.go
@@ -11,6 +11,7 @@ import (
 	"github.com/giantswarm/devctl/v7/cmd/gen/ami"
 	"github.com/giantswarm/devctl/v7/cmd/gen/apptest"
 	"github.com/giantswarm/devctl/v7/cmd/gen/dependabot"
+	"github.com/giantswarm/devctl/v7/cmd/gen/llm"
 	"github.com/giantswarm/devctl/v7/cmd/gen/makefile"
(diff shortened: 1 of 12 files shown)


QuentinBisson:
What do you think about adding smaller rules like follow KISS and do on?

stone-z:
> What do you think about adding smaller rules like follow KISS and do on?

My hope is that these rules can be very minimal and only point to a knowledge base that has all the specific rules. For this hackathon, I am just pointing it to `fmt`, because although some of that content is outdated (operatorkit, etc.) a lot of it still holds, and it might be more cohesive and easier to maintain that than a bunch of rules files. I've been thinking it could use some attention anyway.

HOWEVER, it takes a lot of begging to get cursor to reliably enforce the things there. It _does_ do it, but it might prove not to be reliable enough. If, in the future, we do end up having to distribute the actual rules to every repo, then I would love to have them all be very small, single-focus rule sets (e.g. Shield would maintain one for Go security, one for Helm security, etc.).

fhielpos (approved):
I like it, Picasso.

PR #1423, opened 2025-08-27
Enhance release creation with app dropping and improved UX

njuettner:
Add app dropping support and improve bumpall UX

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `devctl release create`: Drop `karpenter-nodepools` from AWS releases v32.0.0 and higher.
+- `devctl release create --bumpall`: Improve UX by coloring bumped versions green and removed apps red in the summary table.
+- Add support for dropped apps easily.
+
 ## [7.5.3] - 2025-08-26
 
 ### Fixed
--- a/pkg/release/bumpall.go
+++ b/pkg/release/bumpall.go
@@ -17,6 +17,7 @@ import (
 	"github.com/giantswarm/release-operator/v4/api/v1alpha1"
 	"github.com/google/go-github/v74/github"
 	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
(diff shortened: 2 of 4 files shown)


Gacko (approved):
LGTM!

Just one idea, not related to this PR, but when I was using `devctl` to create a Release recently, I was hoping for a way to also define dependencies, maybe as a comma-separated list in the `<app>@<version>` string. So if you got some time, you could also add something like that. ❤️

njuettner:
@Gacko Good idea I added this now, was a quick win 🙂 

I added a description how to do that now dropping, adding or overwriting dependencies:

```
devctl release create -h
Creates and registers a new Giant Swarm platform release including Release CR and release notes.

Usage:
  devctl release create [flags]

Examples:
  # Create a new release by manually specifying new app and component versions
  devctl release create \
        --name v30.1.0 \
        --provider capa \
        --base v30.0.0 \
        --app cert-manager@3.9.2 \
        --component cluster-aws@4.0.1 \
        --overwrite

  # Create a release by bumping all apps and components to their latest versions
  devctl release create \
        --name v30.1.0 \
        --provider capa \
        --base v30.0.0 \
        --bumpall

  # Bump all components, but override a specific app's version and add a dependency
  # This will overwrite any existing dependencies on 'coredns'.
  devctl release create \
        --name v30.1.0 \
        --provider capa \
        --base v30.0.0 \
        --bumpall \
        --app coredns@1.2.3@@cilium

  # Bump all components and remove all dependencies from a specific app
  # This will drop any existing dependencies on 'coredns'.
  devctl release create \
        --name v30.1.0 \
        --provider capa \
        --base v30.0.0 \
        --bumpall \
        --app coredns@1.2.3@@
```

I also added it into the UX:

image

njuettner:
I updated it again for dependencies colors in case we add a remove a dependency otherwise it will not present in green (added) or red (removed)

image

In case we add something:

Example `devctl release create --base v31.1.1 --name v32.0.0 --provider vsphere --bumpall --app cilium@1.2.3@@new-dependency  --overwrite`

image

Gacko:
I was first wondering what's the double `@` for, but then saw you can also already define the component version. Makes sense to me! I think one could discuss about what separator to use for what, like maybe using a format like `app@version(component-version)#dependency,dependency`, but that's nit picking.

PR #1448, opened 2025-09-09
Create branch from parent version tag create-release workflow

mcharriere:
Prior to this change, the branch is created including all the commits
that are meant for the following version.
This change sets the "Release" commit as a base for the new release
branch.

Signed-off-by: Matias Charriere <matias@giantswarm.io>

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Create branch from parent version tag create-release workflow
+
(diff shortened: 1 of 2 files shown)


Gacko:
Random shower thought: From my experience with upstream, I can tell it's actually a better idea to create the branch from the tag, so the moment of releasing. For a v4.0.0 this then would be `release-v4.0.x`. Also this would obsolete the parent logic.

Gacko:
Also it would make sense to not have these branches created upfront, but rather make the creation easier, like automatically resetting a newly created branch `release-v3.6.x` to the latest tag of this minor, so v3.6.1 for example. Because right now these branches pile up completely unused in some repos.

PR #1479, opened 2025-09-16
### Motivation

njuettner:
The `devctl release create --bumpall` command was previously a bit too aggressive, always attempting to update every component to its latest version. This was not ideal for `minor` and `patch` releases, which have stricter rules about what can be changed (e.g., no minor version bumps for Flatcar in a minor release).

This PR makes the command "release-aware" by automatically inferring the release type and applying specific bumping rules, reducing the need for manual intervention and preventing accidental version bumps.

### Changes

-   **Automatic Release Type Inference**: The command now determines if a release is `major`, `minor`, or `patch` by comparing the `--base` and `--name` versions.
-   **Release-Aware Bumping Logic**:
    -   **Major**: Bumps all components and apps to their latest versions.
    -   **Minor**: Restricts `flatcar` to only patch version bumps.
    -   **Patch**: Prevents any version changes for `kubernetes` and `flatcar`, and only allows patch-level bumps for other components and apps.
-   **Provider-Specific Constraints**: The command now reads and enforces version constraints from the `requests.yaml` file located in each provider's directory. This ensures provider-specific requirements are met.
-   **Fixed**: Corrected a table rendering issue that was causing broken lines in the output.

### How to Test

You can test the new logic by running the command with different version increments:

```bash
# Patch Release
./devctl release create --base v31.1.1 --name v31.1.2 --provider aws --bumpall

# Minor Release
./devctl release create --base v31.1.2 --name v31.2.0 --provider aws --bumpall

# Major Release
./devctl release create --base v31.2.0 --name v32.0.0 --provider aws --bumpall
```

Observe the output table and verify that the "DESIRED VERSION" for each component aligns with the new bumping rules.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- `devctl release create --bumpall`:
+  - The release type (`major`, `minor`, `patch`) is now inferred from the `--base` and `--name` versions instead of requiring a separate flag.
+  - The release type is used to enforce version bumping rules for `kubernetes`, `flatcar`, and other components.
+  - Includes version constraints defined in provider-specific `requests.yaml` files.
+
+### Fixed
+
+- Fix a table rendering issue in the `devctl release create --bumpall` output.
+
 ## [7.11.1] - 2025-09-11
 
 ### Changed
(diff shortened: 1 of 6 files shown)


Gacko (approved):

Gacko:
As always: Actually testing this is probably a way better review and since we have more than enough cases to test, I'm sure it's going to work! 🙂

PR #1753, opened 2026-04-14
Fix github repository setup to correctly detect and configure required checks

TheoBrigitte:
This PR fixes an issue where the `devctl repo setup` command was not correctly configuring required check on the default branch.

## Summary

- Switch from `GetCombinedStatus` to `ListCheckRunsForRef` API to correctly detect required status checks when configuring branch protection
- Change default `--checks-filter` from `aliyun` to `create-release` to exclude the release workflow from required checks

## Tests

I've managed to configure checks on the a repo using `devctl repo setup giantswarm/some-repo--log-level debug`

### Before

image

### After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


piontec (commented):
Sorry for taking so long to review. I'm not sure what the problem here is: was repo setup failing for some repos? If yes, then why?

PR #1771, opened 2026-04-23
fix: pr approve-align-files evaluates GitHub Actions check_runs

teemow:
## Problem

`devctl pr approve-align-files` was approving and (via GitHub auto-merge on repos without required status checks) merging Align files PRs that had a failing `pre-commit` GitHub Actions check, as long as the legacy commit statuses (e.g. CircleCI) were green.

Concrete example today: \`giantswarm/mcp-kubernetes#375\` was merged with a failing \`pre-commit\` check because all \`ci/circleci: ...\` checks were \`success\`.

## Root cause

In \`cmd/pr/approvealign/runner.go\`, the \`check_runs\` evaluation loop was gated on \`combinedStatus\` not having a definitive answer:

\`\`\`go
if combinedStatus == nil || (combinedStatus.GetState() != \"success\" && combinedStatus.GetState() != \"failure\") {
    for _, run := range checkRuns.CheckRuns { ... }
}
\`\`\`

GitHub exposes two separate APIs for PR status:

- \`combinedStatus\` (legacy commit statuses) — used by CircleCI (\`ci/circleci: ...\`).
- \`check_runs\` — used by GitHub Actions (e.g. \`pre-commit\`, \`Lint and Test\`).

A green CircleCI made the loop skip entirely, so failing Actions checks were invisible to the tool.

## Fix

Always iterate \`check_runs\` and let any failing run set \`hasFailedChecks = true\` (or any pending run set \`checksPending = true\`). The outer \`!hasFailedChecks\` guard still short-circuits if combinedStatus already detected a failure, so we don't do redundant work.

## Test plan

- [x] \`go build ./cmd/pr/approvealign/\`
- [x] \`go test ./cmd/pr/approvealign/...\` (no tests in this package)
- [ ] Confirmed by re-running \`devctl pr approve-align-files\` against an Align files PR with a failing GitHub Actions check after release — should now report \"Failed checks\" instead of approving.

Made with [Cursor](https://cursor.com)


teemow:
@marians a follow-up of the pre-commit fixes. approve-align-files approved all the PRs although pre-commit was failing for a real reason. This was unexpected for my worflow because my agent would have started agents to fix the real issues. now they all got merged because of auto-merge and pre-commit not being in the list of required_status_checks. this here would be a quick fix to avoid this in the future. but I guess making pre-commit mandatory is the better fix. wdyt?

marians (approved):

PR #1809, opened 2026-05-09
feat: Dev App and Component release support

or4dx:
Added the ability to allow test catalogs to be populated when an app or component uses a dev version(app-version-sha) in its release creation

### Checklist


Gacko (changes requested):
A test version normally comes with a Git commit has at the end, so I'd rather check for that than using SemVer's pre-release part.

njuettner (approved):
LGTM

pkg/release/catalog.go
or4dx, line 1: @Gacko I made the updated.Does this meet this requirement?

Gacko: LGTM! Usually we use the full length SHA, so would be cool to check for a fixed length. We can still amend this in the future, but I'd start as strict as possible.

CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-### Changed
+## [7.40.8] - 2026-05-10
```
Gacko: I don't think this is intended. 🙂

pkg/release/create.go
```diff
@@ -384,6 +384,38 @@ func CreateRelease(name, base, releases, provider string, components, apps []str
 	}
 	newRelease := mergeReleases(effectiveBaseRelease, updatesRelease)
 
+	// Rewrite catalogs to their test variants for apps/components carrying development
+	// (pre-release) versions, so the release-operator can locate the build artifact.
```
Gacko: Just a nit and mostly for your knowledge: There's no Release Operator anymore. We are consuming these CRs in the `cluster` chart via lookup and in some other places, but we no longer run the Release Operator. This is why the `sdk` module in the `releases` repo is the single source of truth for this CRD by now.

Gacko (changes requested):

pkg/release/catalog.go
or4dx: Noted. I made the update to use exactly 40 characters as the criteria for a valid dev version.

pkg/release/create.go
```diff
@@ -384,6 +384,38 @@ func CreateRelease(name, base, releases, provider string, components, apps []str
...
+			newRelease.Spec.Apps[i].Catalog = testCatalog
+		}
+	}
+	for i, comp := range newRelease.Spec.Components {
+		if isDevVersion(comp.Version) {
+			// Components with no explicit catalog (e.g. kubernetes, flatcar, os-tooling)
+			// are used for AMI name lookup, not helm chart deployment — no catalog change needed.
+			if comp.Catalog == "" {
+				continue
+			}
```
Gacko, line 411: `cluster-<provider>` is a Helm chart, but normally references `cluster` catalog. Flatcar, Kubernetes and OS Tooling are just there for the version, there's no catalog indeed.

So from a pure "how the CRD is intended to work" point of view, I'd still use `default` catalog then and render it like that.

On the other hand you're right, in reality we currently do not have such cases and it would rather confuse people to have OS Tooling rendered with `default-test` catalog in the end.

So yeah, implementation fine, but please update your comment as it seems a bit misleading to me. 🙂

or4dx: @Gacko Comment updated.

Gacko (approved):

PR #1832, opened 2026-05-21
feat: add devctl repo checks --update command

QuentinBisson:
## What

Adds `devctl repo checks --update --checks <list> REPOSITORY` — a surgical command that unions named checks into the default branch's required status checks without touching any other branch protection or repo settings.

The existing `devctl repo setup` rewrites permissions, merge settings, renovate config, etc. This command only PATCHes `required_status_checks`, leaving everything else untouched.

Also renames the `Values and schema` workflow name and its `check` job to `check-values-schema` so the status check name is unambiguous in branch protection rules and PR status views.

## Why

The align-files sync (`giantswarm/github`) needs to ensure conventional-commit and other standard checks are required on every repo's default branch. The `check` job name was too generic to use as a required check reference.

## Behaviour

- `--update` reads current required checks via `GET .../protection/required_status_checks`, unions in `--checks`, and PATCHes back
- If the branch has no protection configured, the command warns and skips rather than failing
- Checks already present are deduplicated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- `devctl repo checks --update --checks <list> REPOSITORY` adds named checks to the required status checks on the default branch protection rule without touching any other branch protection or repo settings.
+- Rename the generated `Values and schema` workflow name and its `check` job to `check-values-schema` for an unambiguous required-check reference in branch protection rules.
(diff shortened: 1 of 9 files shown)


paurosello (approved):
Approved. The `repo checks --update` command is surgical and bounded — PATCH for the common case, fall back to a full `UpdateBranchProtection` only when required_status_checks isn't configured yet, warn-and-skip when there's no protection at all. The `getGithubChecks` improvement (reading GH-Actions check runs in addition to legacy commit statuses) is a separate concern but net-additive and called out in the body.

Two small flags for follow-up, not blocking:
- The `Values and schema` → `check-values-schema` rename is a soft-breaking change for any repo already requiring the old job name in branch protection — until align-files re-syncs each repo (giantswarm/github#5269), they'll have a stale required-check name. Known transition cost worth surfacing in the rollout.
- Consider splitting the `getGithubChecks` change into its own PR next time for cleaner bisecting.

PR #1833, opened 2026-05-21
feat(gen/workflows): route security: commits to ### Security in release-please-config

QuentinBisson:
## Summary

Aligns `devctl gen workflows`' generated `release-please-config.json` with [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) across both `--changelog-style` modes and adds a Giant-Swarm-specific `security:` conventional commit type that routes to `### Security`.

- Both modes now include `{"type": "security", "section": "### Security"}` in `changelog-sections`.
- `--changelog-style=release-please` no longer falls back to Release Please's Angular defaults (`### Features` / `### Bug Fixes` / `### Other Changes`). It now emits the full Angular-to-KaC mapping: `feat` to `### Added`, `fix` to `### Fixed`, and the rest of the Angular types (`perf`, `revert`, `refactor`, `docs`, `style`, `test`, `build`, `ci`, `chore`) to `### Changed`.
- `--changelog-style=legacy` is unchanged except for the added `security` entry.
- No `hidden: true` on any type. Release Please already excludes its own `chore(release):` PRs from the generated CHANGELOG.

## Why `security:` rather than `fix(security):`

Release Please's `changelog-sections` matches on the conventional commit type only. Scopes (`fix(security):`) cannot be routed to a non-`Fixed` section. Under fully-automated release-please flows, where the release PR auto-merges without human editing, a dedicated `security:` type is the only machine-driven path to `### Security`.

This aligns with the same `CC -> KaC` mapping used in the team-bumblebee Slack digest (giantswarm/github PR #5270), so changelogs and weekly digests stay consistent.

## Scope

In scope:

- Template + README + CHANGELOG in `devctl`.
- Unit tests asserting both modes render valid JSON with the expected `type -> section` map.

Out of scope:

- Migrating existing repos' `release-please-config.json` (separate per-repo PRs).
- Deprecating `--changelog-style=legacy`.
- Adding a `### Maintenance` section (not part of KaC).
- Detecting `BREAKING CHANGE:` footers in PR bodies (Release Please already handles `!` in titles).

After this lands plus the matching change in `giantswarm/github-workflows`'s `validate-changelog.yaml`, run `devctl gen workflows` across team-bumblebee repos to regenerate.


paurosello (approved):
Approved. The "why `security:`-as-type, not `fix(security):`" reasoning is right — release-please's `changelog-sections` matches on type only; scopes don't participate. There's no other path to `### Security` under fully-automated release-please flow.

Making `--changelog-style=release-please` emit the full Angular→KaC mapping explicitly (instead of falling back to release-please's hidden Angular defaults) is the right call — turns an implicit contract into an explicit one. Tests target JSON shape directly via the template, which is the right unit. Both modes covered.

Pairs cleanly with #5270's CC→KaC table in the team digest, so generated CHANGELOG.md and the Slack digest speak the same language.

fiunchinho (approved):

PR #1953, opened 2026-06-11
feat(gen/renovate): add reviewers subcommand to edit existing config

marians:
## Goal

Allow setting the `reviewers` for Renovate's pull requests in a repo's config, so teams can reduce notification noise by directing Renovate review requests to specific people/teams.

## What

Two complementary entry points:

**1. `gen renovate --reviewers` (generator flag)** — bakes a top-level `reviewers` array into the generated `renovate.json5`.

```sh
devctl gen renovate --language go --reviewers team:team-rocket
```

This is the single source of truth for repos whose config is regenerated by align-files (`devctl gen renovate`, see #1919): because reviewers live in the template, they survive regeneration. align-files in `giantswarm/github` can wire it from per-repo config, the same way `--circleci-generated` is.

**2. `gen renovate reviewers` (surgical subcommand)** — sets the top-level `reviewers` array in an **existing** Renovate config (`renovate.json5` preferred, else `renovate.json`) in the current directory. For hand-maintained (non-generated) configs.

```sh
devctl gen renovate reviewers -r user1 -r user2
```

## How

**Generator flag:** `--reviewers`/`-r` (repeatable, comma-separated) is threaded through the generator config → params → template, mirroring `--circleci-generated`. Empty omits the key.

**Surgical subcommand:**
- **In-place edit.** A small comment- and string-aware JSON5 scanner (`pkg/renovate`) locates the *top-level* `reviewers` key (or inserts it as the first key if absent) and rewrites only its value. All comments, key order, quoting and formatting elsewhere are preserved byte-for-byte — no parse-and-reserialize, so hand-maintained configs aren't reformatted.
- **Quote & indent follow the file**, not the extension: the dominant of single/double quotes among *top-level value* strings wins (nested `customManagers` regex can't sway it; extension is only a fallback for empty/quote-less files). The inserted key's quote follows the first root key's own style, and the existing member indentation (tabs / N spaces) is reused.
- **Atomic write:** writes to a temp file + `os.Rename`, preserving file mode, so an interrupted write can't truncate the original.
- **Edge cases:** errors if no Renovate config is found; `Args: cobra.NoArgs` rejects stray positionals; replaces an existing `reviewers` value; empty `renovate.json` insertion stays valid strict JSON (no dangling trailing comma).

## Testing

- **Generator:** tests assert reviewers are omitted by default and otherwise rendered as a valid, double-quoted top-level array (re-parsed with a JSON5 parser), including combined with `--circleci-generated` + `--interval`.
- **Subcommand:** hand-crafted table tests covering insertion/replacement, nested `additionalReviewers`/`reviewersFromCodeOwners` traps, comments between key and value, mixed/nested quotes, key-vs-value quote styles, tab/4-space indentation, file-mode preservation, and strict-JSON validity.
- **Corpus of 16 real giantswarm Renovate configs** (`testdata/wild/`), each independently re-parsed with a third-party JSON5 parser to assert the edit stays valid, sets reviewers correctly, preserves every other field and all comments, and matches the file's own quote style.

Adds `github.com/titanous/json5` as a test-only dependency (validation oracle).

### Checklist


pkg/renovate/reviewers.go
```diff
@@ -0,0 +1,156 @@
...
+		key = string(quote) + "reviewers" + string(quote)
+	}
+
+	// When the root object already has members, a trailing comma after our
+	// entry is just the separator before the existing first key and is valid
+	// everywhere. When the object is empty, that comma would be a dangling
+	// trailing comma which strict JSON (renovate.json) rejects, so omit it.
+	var insertion string
+	if rootObjectIsEmpty(src, objStart) {
+		insertion = "\n  " + key + ": " + array + "\n"
```
marians: **Inserted key hardcodes a 2-space indent, ignoring the file’s actual indentation.**

The insert path always prefixes `"\n  "` (2 spaces), but the package doc promises to preserve `formatting`. For a config indented with tabs or 4 spaces, the new `reviewers` line is misaligned against every sibling key:

```jsonc
{
	"extends": ["foo"]   // tab-indented file
}
```
becomes
```jsonc
{
  "reviewers": ["team:x"],
	"extends": ["foo"]
}
```
Still valid, but inconsistent. Consider detecting the leading whitespace of the existing first member (or the brace’s line) and reusing it. All wild fixtures appear to be 2-space, so this path is untested for other styles.

pkg/renovate/reviewers.go
```diff
@@ -0,0 +1,156 @@
...
+		out = append(out, src[span.valueEnd:]...)
+		return out, nil
+	}
+
+	// Insert `reviewers: [...]` as the first key of the root object, right
+	// after the opening brace, matching the file's existing key-quoting style
+	// (bare identifier vs quoted string).
+	key := "reviewers"
+	if rootKeysAreQuoted(src, objStart, !defaultSingle) {
+		key = string(quote) + "reviewers" + string(quote)
```
marians: **Inserted key is quoted with the dominant *value* quote, not the file's *key* quote style.**

`quote` here comes from `detectQuoteStyle`, which counts all string literals (values included). But this is the *key* being wrapped. In a `.json5` file with double-quoted keys whose string values happen to be mostly single-quoted, e.g.

```json5
{
  "extends": ['foo'],
  "x": ['a', 'b'],
}
```
single values outnumber double, so `quote == '\''` and the inserted key becomes `'reviewers'` — a single-quoted key amid double-quoted keys. Still valid JSON5, but inconsistent with the surrounding key style this block claims to match. The key quote should be derived from the first root key's own quote char, not from the value-quote heuristic.

pkg/renovate/scanner.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+		return '"'
+	}
+}
+
+// countStringQuotes counts how many string literals are opened with a single
+// vs a double quote. It is comment-aware and skips the interior of each string,
+// so quote characters appearing inside another string or a comment (a common
+// case in Renovate regex managers) are not counted.
+func countStringQuotes(src []byte) (single, double int) {
```
marians: **Dominant-quote heuristic can pick double quotes for a single-quote house-style `.json5`.**

`countStringQuotes` weights every string literal equally. A house-style `renovate.json5` (single-quoted `extends`/`reviewers`) that also carries a `customManagers` block with several double-quoted `matchStrings`/regex entries can easily have `double > single`. `detectQuoteStyle` then returns `"`, and `reviewers` is rendered double-quoted into an otherwise single-quoted file — valid, but contradicting the documented house-style intent. The `quoteOfExtends` oracle in the wild test would also disagree for such a file; none of the harvested fixtures trigger it, so the case is latent/untested. Consider biasing toward the quote used by top-level (root) string values rather than a flat count that regex managers can dominate.

pkg/renovate/reviewers.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	defaultSingle := strings.EqualFold(filepath.Ext(path), ".json5")
+
+	out, err := setReviewers(src, reviewers, defaultSingle)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = os.WriteFile(path, out, info.Mode())
```
marians: **In-place write is non-atomic — an interrupted write can corrupt the very file this tool exists to preserve.**

`os.WriteFile` truncates `path` and then writes `out`. If the process is killed, the disk fills, or the write errors partway through, the user's existing `renovate.json5` is left truncated/empty with no backup. Given the whole point of this package is byte-for-byte preservation, a write-to-temp-then-`os.Rename` (atomic replace) would be safer. (Minor, related: the separate `os.Stat` + `info.Mode()` is effectively dead — `os.WriteFile` only applies the perm arg when *creating* a file, so for an existing file the mode argument is ignored.)

cmd/gen/renovate/reviewers/command.go
```diff
@@ -0,0 +1,62 @@
...
+	f := &flag{}
+
+	r := &runner{
+		flag:   f,
+		logger: config.Logger,
+		stderr: config.Stderr,
+		stdout: config.Stdout,
+	}
+
+	c := &cobra.Command{
```
marians, line 56: **No `Args` constraint — positional arguments are silently ignored.**

The command takes reviewers only via `--reviewers/-r`. A user who runs `devctl gen renovate reviewers team:team-rocket` (positional, no flag) has the argument silently discarded and then hits the confusing `--reviewers must not be empty` error. Setting `Args: cobra.NoArgs` on the command would reject stray positional args with a clear message and guide users to the flag.

pkg/renovate/reviewers.go
marians: Fixed in edf815b. The insert path now reuses the existing members' leading whitespace (`lineIndent` captures the first member's indent — tabs or N spaces — and falls back to 2 spaces only for empty/single-line objects). Added tab and 4-space table tests.

pkg/renovate/reviewers.go
marians: Fixed in edf815b. The inserted key is now quoted with the *first root key's own* quote char (`rootStyle.keyQuote`), decoupled from the value-quote vote. Added a test with double-quoted keys + single-quoted values asserting `"reviewers": ['r']`.

pkg/renovate/scanner.go
marians: Fixed in edf815b. `analyzeRootStyle` now counts only string literals shallow within each top-level value (depth ≤ 1 via `addShallowStringQuotes`), so `extends`/`reviewers` elements vote but nested `customManagers`/`packageRules` regex do not. Added a test for a single-quote house style with double-quoted nested matchStrings.

pkg/renovate/reviewers.go
marians: Fixed in edf815b. Writes now go through `writeFileAtomic` (temp file in the same dir + `os.Rename`), so an interrupted write can't truncate the original. The mode is applied to the temp file via `Chmod` before the rename (so the `Stat` is no longer dead), and a mode-preservation test was added.

cmd/gen/renovate/reviewers/command.go
marians: Fixed in edf815b — added `Args: cobra.NoArgs`. A stray positional now errors with a clear `Unknown command "…"` message instead of being silently ignored.

teemow (commented):
Code review of the post-fixup state (edf815b): the surgical editor is solid. I re-ran the test suite locally and additionally probed CRLF line endings, trailing commas after the last member, block comments between brace and first key, single-line root objects, and the exact output of `gen renovate --circleci-generated` — all edits stay valid and style-consistent. The atomic write, root-style detection, and `NoArgs` fixes all look right.

One architectural gap you likely didn't know about — #1919 / align-files interaction:

Since #1919, the align-files action in `giantswarm/github` runs `devctl gen renovate --language <lang> --circleci-generated` for every repo with `gen.ci.generate: true`. That is a **full regeneration from the template**, not a surgical edit — so on those repos, any `reviewers` set by this new subcommand is silently wiped on the next align-files sync. The template has no `reviewers` support.

Today the `ci.generate` rollout covers only a handful of repos, so the subcommand works as intended almost everywhere — but the overlap will grow. Suggestions (follow-up PRs are fine, doesn't need to block this one):

1. Add `Reviewers` to the `gen renovate` generator config/template too, so generated repos can carry them through regeneration (wired from per-repo config in `giantswarm/github`, like `--circleci-generated` is).
2. Alternatively, have align-files run `gen renovate reviewers` *after* `gen renovate` for repos that declare reviewers.
3. At minimum, mention in the command's long help that repos managed by align-files CircleCI generation will have this setting overwritten until the generator learns about it.

Option 1 keeps a single source of truth (the repo config in giantswarm/github) and avoids ordering subtleties, so I'd lean that way.

teemow (approved):
Verified 3394b05 + the main merge locally: build green after `go generate`, full renovate test suites pass, and an end-to-end smoke run confirms `gen renovate -l go -r team:a -r team:b --circleci-generated -i weekly` renders a valid JSON5 with the top-level `reviewers` array, and `gen renovate reviewers` still surgically replaces that value in the generated file without touching anything else. The help-text pointer for align-files-managed repos reads well, and `--reviewers` (plural) matching `gen dependabot` is the right call.

This resolves the #1919 overlap — align-files just needs the follow-up wiring in giantswarm/github to pass per-repo reviewers. Approving.

PR #1971, opened 2026-06-12
feat(gen renovate): emit idiomatic JSON5 formatting

marians:
### Checklist

## What

Switches the `renovate.json5` produced by `devctl gen renovate` to JSON5-native formatting:

- **unquoted keys** (`extends:`, `packageRules:`, `enabled:`, …)
- **single-quoted values** (`'github>giantswarm/...'`, `'orb'`, …)
- **one array item per line, with trailing commas** (the previously inline arrays like `["orb"]` and `schedule` are expanded)

## Why

The output went into a `.json5` file but was written as pure quoted JSON, paying the non-standard-format cost without using any of the JSON5 readability features. This style now matches both devctl's own hand-maintained `renovate.json5` and the `giantswarm/renovate-presets`, so the preset, generated, and `renovate-custom.json5` layers a reader sees are visually uniform.

The one-item-per-line rule for arrays in combination with trailing commas ensure minimal, readable diffs when adding or removing items.

This is **cosmetic only** — no change to Renovate's behavior. Because it is align-files-synced, repos will see a one-time formatting diff on their next sync.

## Tests

- Added a **golden-file test** (`Test_Golden`, with `-update` support) that pins the exact rendered output byte-for-byte for representative configs, so future template edits can't silently drift the formatting. Each golden is also round-tripped through a JSON5 parser to prove validity.
- Added fixtures covering the previously **untested `python` and `schedule` branches**, plus `default-go` and a `full` (all options on) case.
- Updated the existing substring assertions to the new quote style.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/gen/input/renovate/internal/file/renovate.json5.template
```diff
@@ -4,44 +4,50 @@
...
-    "github>giantswarm/{{ .RepoName }}:renovate-custom.json5",
+    'github>giantswarm/{{ .RepoName }}:renovate-custom.json5',
     {{- end }}
   ],
   {{- if .Reviewers }}
-  "reviewers": [
+  reviewers: [
     {{- range .Reviewers }}
-    {{ . | printf "%q" }},
+    {{ . | printf "'%s'" }},
```
marians: **Removed escaping: `printf "'%s'"` no longer escapes the value.** This line (and `schedule:` at line 50) interpolate a free-text flag value directly into a JSON5 single-quoted string. The previous `printf "%q"` produced a properly-escaped, always-valid quoted string; the new form emits the raw value verbatim.

`--reviewers` and `--interval` are unvalidated free text (`flag.Validate()` only checks `Language`), so any value containing a `'`, a `\`, or a newline now produces malformed JSON5 — e.g. `--interval "before 5am 'monday'"` renders `schedule: [ 'before 5am 'monday'' ]`, which fails to parse and breaks the generated file. Realistic reviewer/schedule values rarely contain these characters, so the practical risk is low, but the change does trade away a robustness guarantee that `%q` gave for free.

If you want to keep the single-quote style while staying safe, escape embedded quotes/backslashes before emitting (e.g. a small template func that backslash-escapes `'` and `\`).

marians: Good catch — fixed in b4a6fe0.

I kept the single-quote style but restored the safety `%q` gave us, doing the escaping in the input layer rather than the template (text/template's `printf` has no single-quote-escaping verb, and the shared executor registers no FuncMap). A `squote` helper escapes `'`, `\` and control chars, and the template now emits the pre-quoted values verbatim.

As you note, reviewer/team values are GitHub slugs so they were never at risk — but `--interval` is free text, so it's now escaped defensively. `--interval "before 5am 'monday'"` renders `'before 5am \'monday\''` and parses fine.

The goldens are unchanged (the escaping is a no-op for normal values), and I added `Test_FreeTextValuesEscaped` with a hostile interval (`' \ ` + newline) asserting the output stays valid JSON5 and round-trips to the original value.

QuentinBisson (approved):

fiunchinho (approved):

PR #2096, opened 2026-07-23
fix(gen precommit): run the whole helm schema pipeline as one hook

piontec:
Fixes giantswarm/giantswarm#37267 (and giantswarm/giantswarm#37234).
Tested with `hello-world-app` to cover the followint test cases:
- default `values.yaml` validation
- extended `values.yaml` with values matching a schema imported with `$ref`
- extended `values.yaml` with values breaking a schema imported with `$ref`

## What

Each chart's `values.schema.json` is now generated, fixed and normalized by a **single** `repo: local` pipeline hook:

```
helm schema  ->  rewrite $ref siblings  ->  schemalint normalize
```

This replaces **both** the external `losisin/helm-values-schema-json` hook and the standalone `schemalint-normalize` hook. Read-only `schemalint-verify` stays as-is.

## Why: two independent reasons `pre-commit run -a` could never converge

pre-commit does **not** feed hooks into each other. It runs each hook independently against the file on disk and fails the run if any hook modified it. For a multi-fixer setup to pass, every fixer must agree on **one** resting format (a fixed point).

### 1. Rival formatters — no fixed point (#37267)

Three fixers wanted different resting formats for the same file:

| fixer | resting format |
|---|---|
| `helm schema` | regenerates wholesale in the **generator's** key ordering, discarding existing formatting |
| `schemalint normalize` | rewrites into **schemalint's** canonical ordering (+ unicode escaping) |
| the `$ref` fix | rewrites the generator's output again |

Measured on `hello-world`: the generator's form and schemalint's normalized form differ by **2189 diff lines** (pure key reordering — semantically identical). So committing either one made the other hook fail, and reruns ping-ponged A → B → A → B forever. Exactly one hook always failed; **no committed file could make CI green**, blocking the `pre-commit` check on every chart repo that got the latest align-files PR.

This surfaced only after #2098. That PR correctly made `schemalint-normalize` actually run (before it silently no-op'd on `accepts 1 arg(s)`), which exposed the pre-existing incompatibility. #2098 stays — `main` only looked green because normalization was never enforced.

**Fix:** chaining the steps in one hook makes `schemalint normalize` always the **last writer**, so the normalized form is the single resting format and the committed schema is a fixed point. `schemalint-verify` never rewrites, so it was never part of the conflict.

### 2. Invalid schema next to `$ref` (#37234, upstream losisin/helm-values-schema-json#317)

Under `noAdditionalProperties: true` the generator emits `additionalProperties: false` as a sibling of every bare `$ref`. Per JSON Schema 2020-12 (§10.3.2) `additionalProperties` only considers **sibling** `properties`/`patternProperties` — never properties pulled in through `$ref` — so it rejects every field the referenced schema defines:

```
$ helm template test helm/hello-world -f values-ext.yaml
Error: ... at '/affinity': additional properties 'podAntiAffinity' not allowed
```

The correct keyword is `unevaluatedProperties`, which the tool cannot emit and won't substitute. Newer Helm (3.19.x, via `santhosh-tekuri/jsonschema/v6`) enforces 2020-12 strictly, so this now breaks `helm template`. The hook rewrites those nodes to `unevaluatedProperties: false`.

## Tooling: nothing new to install

`schemalint` is installed by the hook itself via `additional_dependencies` (hence `language: golang`) — verified that pre-commit installs it and puts it on `PATH`, so local dev keeps auto-install exactly as before. `helm schema` needs the `schema` helm plugin, already installed by the generated pre-commit CI workflow (`pre-commit-action.yaml.template`, gated on `HasHelmchart`) — and the same requirement the external hook had, since it just ran `helm schema`.

**Renovate:** a new custom manager tracks the `additional_dependencies` pin with `depName: giantswarm/schemalint`, the same depName as the `schemalint-verify` `rev:`. Renovate consolidates both into one update, so the normalizing binary can never drift from the verifying one — a version skew between them could reintroduce the rival-formatter conflict.

## Changes

- `pre-commit-config.yaml.template` — one combined pipeline hook; removed the external generator hook and the standalone `schemalint-normalize`.
- `precommit_test.go` — asserts the single-hook shape, the step order (`normalize` last), `additional_dependencies` pin, that `schemalint-verify` runs after the pipeline, and that **none** of the three steps reappears as a separate hook.
- `renovate-custom.json5` — manager for the new pin.
- `CHANGELOG.md`.

## Verification

- `go build ./...`, `go vet`, `gofmt`, `go test ./pkg/gen/input/precommit/...` — green.
- Reproduced the conflict first: generator form vs normalized form differ (2189 lines); normalized form **is** a fixed point of `normalize`; `normalize` preserves all 9 `unevaluatedProperties`.
- Built `devctl` from this branch → `devctl gen precommit --flavors helmchart` on the real `hello-world` chart → generated config has exactly one pipeline hook + `schemalint-verify`.
- `pre-commit run -a`: run 1 settles the schema to the normalized form; **runs 2 and 3 pass every hook with exit 0**, byte-identical schema, `git status` clean. This is the exact scenario from #37267's reproduction that previously never converged.
- `helm template` with valid extended affinity → **succeeds**; with a bogus key → **rejected** (`additional properties 'frog' not allowed`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
+++ b/pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
@@ -139,8 +139,10 @@ repos:
   # This replaces both the external `losisin/helm-values-schema-json` hook and the
   # separate `schemalint-normalize` hook. `schemalint` is installed by the hook itself
   # (`additional_dependencies`, hence `language: golang`); `helm schema` needs the `schema`
-  # helm plugin, installed in the pre-commit CI workflow -- the same requirement the
-  # external hook had, since it just ran `helm schema`.
+  # helm plugin -- the same requirement the external hook had, installed in the pre-commit
+  # CI workflow. The external hook wrapped `helm schema` in a script that checked for the
+  # plugin first, so the leading `helm plugin list` guard below keeps that actionable
+  # message (a dev without the plugin would otherwise just get `unknown command "schema"`).
   - repo: local
     hooks:
       - id: helm-schema-{{.}}
@@ -152,7 +154,7 @@ repos:
         require_serial: true
         entry: sh -c
         args:
-          - 'helm schema --config helm/{{.}}/.schema.yaml && python3 -c ''import json,sys; h=lambda o: ({**{k:v for k,v in o.items() if k!="additionalProperties"},"unevaluatedProperties":False}) if ("$ref" in o and o.get("additionalProperties") is False) else o; p=sys.argv[1]; d=json.load(open(p),object_hook=h); f=open(p,"w"); json.dump(d,f,indent=4); f.write(chr(10))'' helm/{{.}}/values.schema.json && schemalint normalize helm/{{.}}/values.schema.json -o helm/{{.}}/values.schema.json --force'
+          - 'helm plugin list | grep -q "^schema[[:space:]]" || { echo "helm schema plugin missing, install it with: helm plugin install https://github.com/losisin/helm-values-schema-json"; exit 1; }; helm schema --config helm/{{.}}/.schema.yaml && python3 -c ''{{ $.RefFixPython }}'' helm/{{.}}/values.schema.json && schemalint normalize helm/{{.}}/values.schema.json -o helm/{{.}}/values.schema.json --force'
   - repo: https://github.com/giantswarm/schemalint
     rev: v2.6.1
     hooks:
--- a/pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
+++ b/pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
@@ -103,24 +103,59 @@ repos:
 {{- if .HasHelmchart }}
 {{- range .HelmCharts }}
 
-  # Helm chart hooks for {{.}}
-  - repo: https://github.com/losisin/helm-values-schema-json
-    rev: v2.5.0
+  # Helm chart schema for {{.}}: generate -> fix $ref -> normalize, as ONE hook.
+  #
+  # This single local hook deliberately runs the whole schema pipeline. pre-commit does
+  # NOT feed hooks into each other: it runs each hook independently against the file on
+  # disk and fails the run if any hook modified it. So for a multi-fixer setup to pass,
+  # all fixers must agree on ONE resting format. Splitting these steps into separate
+  # hooks gives them rival resting formats and `pre-commit run -a` can then never
+  # converge -- reruns just ping-pong between formats and one hook always fails:
+  #
+  #   * `helm schema` regenerates the schema from values.yaml wholesale, in the
+  #     generator's own key ordering, and does not preserve existing formatting.
+  #   * `schemalint normalize` rewrites it into schemalint's canonical key ordering
+  #     (and unicode escaping), which is NOT the generator's ordering.
+  #   * the $ref fix below rewrites the generator's output again.
+  #
+  # Chained here, `schemalint normalize` is always the LAST writer, so the normalized
+  # form is the single resting format and the committed file is a fixed point -- the
+  # hook is a no-op on an already-processed schema and the run is clean. Read-only
+  # `schemalint-verify` then checks that resting format (it never rewrites, so it is not
+  # part of the conflict). See giantswarm/giantswarm#37267.
+  #
+  # The `$ref` step works around a generator bug: with `noAdditionalProperties: true` it
+  # emits `additionalProperties: false` as a sibling of every bare `$ref`. Per JSON Schema
+  # 2020-12, `additionalProperties` only sees sibling `properties`/`patternProperties`
+  # -- NOT properties pulled in through `$ref` -- so it wrongly rejects every field the
+  # referenced schema defines (e.g. `helm template` fails with "additional properties
+  # 'podAntiAffinity' not allowed" for a k8s Affinity). The correct 2020-12 keyword is
+  # `unevaluatedProperties`, which DOES consider $ref-evaluated properties, but the tool
+  # cannot emit it and does not drop the bogus `additionalProperties` when annotated. So
+  # for every object that has both `$ref` and `additionalProperties: false` we drop
+  # `additionalProperties` and set `unevaluatedProperties: false`.
+  # Upstream bug: https://github.com/losisin/helm-values-schema-json/issues/317
+  #
+  # This replaces both the external `losisin/helm-values-schema-json` hook and the
+  # separate `schemalint-normalize` hook. `schemalint` is installed by the hook itself
+  # (`additional_dependencies`, hence `language: golang`); `helm schema` needs the `schema`
+  # helm plugin, installed in the pre-commit CI workflow -- the same requirement the
+  # external hook had, since it just ran `helm schema`.
+  - repo: local
     hooks:
-      - id: helm-schema
+      - id: helm-schema-{{.}}
+        name: "generate + fix + normalize helm/{{.}}/values.schema.json"
+        language: golang
+        additional_dependencies: ['github.com/giantswarm/schemalint/v2@v2.6.1']
+        files: ^helm/{{.}}/(values\.yaml|\.schema\.yaml|values\.schema\.json)$
+        pass_filenames: false
+        require_serial: true
+        entry: sh -c
         args:
-          - --config
-          - helm/{{.}}/.schema.yaml
+          - 'helm schema --config helm/{{.}}/.schema.yaml && python3 -c ''import json,sys; h=lambda o: ({**{k:v for k,v in o.items() if k!="additionalProperties"},"unevaluatedProperties":False}) if ("$ref" in o and o.get("additionalProperties") is False) else o; p=sys.argv[1]; d=json.load(open(p),object_hook=h); f=open(p,"w"); json.dump(d,f,indent=4); f.write(chr(10))'' helm/{{.}}/values.schema.json && schemalint normalize helm/{{.}}/values.schema.json -o helm/{{.}}/values.schema.json --force'
   - repo: https://github.com/giantswarm/schemalint
     rev: v2.6.1
     hooks:
-      - id: schemalint-normalize
-        pass_filenames: false
-        args:
-          - helm/{{.}}/values.schema.json
-          - -o
-          - helm/{{.}}/values.schema.json
-          - --force
       - id: schemalint-verify
         pass_filenames: false
         args:
--- a/pkg/gen/input/precommit/precommit_test.go
+++ b/pkg/gen/input/precommit/precommit_test.go
@@ -98,6 +98,102 @@ func Test_New_WithHelmchartFlavor(t *testing.T) {
 	}
 }
 
+// Test_HelmSchemaFixHook verifies the whole schema pipeline is emitted as a SINGLE local
+// hook per chart: `helm schema` generate -> $ref fix (additionalProperties:false ->
+// unevaluatedProperties:false, losisin/helm-values-schema-json#317) -> `schemalint
+// normalize`.
+//
+// It must be one hook. pre-commit runs each hook independently against the file on disk
+// and fails the run if any hook modified it, so every fixer must agree on one resting
+// format. As separate hooks these three have rival resting formats (the generator's key
+// ordering vs schemalint's canonical ordering) and `pre-commit run -a` can never converge
+// (giantswarm/giantswarm#37267). Chaining them with `schemalint normalize` LAST makes the
+// normalized form the single resting format, so the assertions below pin both the
+// single-hook shape and the step order.
+func Test_HelmSchemaFixHook(t *testing.T) {
+	dir := t.TempDir()
+
+	chartDir := filepath.Join(dir, "helm", "test-chart")
+	if err := os.MkdirAll(chartDir, 0755); err != nil {
+		t.Fatalf("setup failed: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(chartDir, "Chart.yaml"), []byte("name: test-chart\n"), 0644); err != nil {
+		t.Fatalf("setup failed: %v", err)
+	}
+
+	origDir, _ := os.Getwd()
+	if err := os.Chdir(dir); err != nil {
+		t.Fatalf("chdir failed: %v", err)
+	}
+	defer func() { _ = os.Chdir(origDir) }()
+
+	got := renderConfig(t, Config{
+		Language: "go",
+		Flavors:  []string{"helmchart"},
+		RepoName: "my-repo",
+	})
+
+	for _, want := range []string{
+		"id: helm-schema-test-chart",
+		"helm schema --config helm/test-chart/.schema.yaml",
+		"unevaluatedProperties",
+		"helm-values-schema-json/issues/317",
+		"schemalint normalize helm/test-chart/values.schema.json",
+		// schemalint is installed by the hook itself, so no new tooling is required.
+		"additional_dependencies: ['github.com/giantswarm/schemalint/v2@v2.6.1']",
+		"language: golang",
+	} {
+		if !strings.Contains(got, want) {
+			t.Errorf("expected %q in rendered config, got:\n%s", want, got)
+		}
+	}
+
+	// No step of the pipeline may be a separate hook: not the external generator hook,
+	// not the old standalone $ref-fix hook, and not a standalone schemalint-normalize.
+	// Any of those reintroduces rival resting formats (see the function doc). The
+	// generator's `repo:`+`rev:` line pair is matched rather than the bare URL, because
+	// the explanatory comment legitimately mentions the URL in prose.
+	for _, unwanted := range []string{
+		"helm-values-schema-json\n    rev:",
+		"id: fix-schema-ref-unevaluated-test-chart",
+		"id: schemalint-normalize",
+	} {
+		if strings.Contains(got, unwanted) {
+			t.Errorf("did not expect %q in rendered config (the whole schema pipeline must be one local hook), got:\n%s", unwanted, got)
+		}
+	}
+
+	// Step order INSIDE the single hook's command: generate -> $ref fix -> normalize.
+	// `schemalint normalize` must be the LAST writer, otherwise the resting format is not
+	// the normalized one and the run never converges. Asserted on the command line itself,
+	// not the whole config, since the explanatory comment mentions the same tool names.
+	var pipeline string
+	for _, line := range strings.Split(got, "\n") {
+		if strings.Contains(line, "helm schema --config") {
+			pipeline = line
+			break
+		}
+	}
+	if pipeline == "" {
+		t.Fatalf("no pipeline command line found in rendered config:\n%s", got)
+	}
+	genIdx := strings.Index(pipeline, "helm schema --config")
+	fixIdx := strings.Index(pipeline, "unevaluatedProperties")
+	normIdx := strings.Index(pipeline, "schemalint normalize")
+	if !(genIdx >= 0 && genIdx < fixIdx && fixIdx < normIdx) {
+		t.Errorf("pipeline order must be generate -> $ref fix -> normalize (normalize last); "+
+			"got positions generate=%d, fix=%d, normalize=%d in:\n%s", genIdx, fixIdx, normIdx, pipeline)
+	}
+
+	// Read-only verify still runs, and only after the pipeline produced the resting format.
+	hookIdx := strings.Index(got, "id: helm-schema-test-chart")
+	verifyIdx := strings.Index(got, "id: schemalint-verify")
+	if !(hookIdx >= 0 && hookIdx < verifyIdx) {
+		t.Errorf("schemalint-verify must run after the pipeline hook; got pipeline=%d, verify=%d in:\n%s",
+			hookIdx, verifyIdx, got)
+	}
+}
+
 func Test_New_WithoutHelmchartFlavor(t *testing.T) {
 	p, err := New(Config{
 		Language: "go",
--- a/renovate-custom.json5
+++ b/renovate-custom.json5
@@ -64,18 +64,26 @@
       datasourceTemplate: 'github-tags',
     },
     // schemalint pinned in `additional_dependencies` of the local helm-schema pipeline
-    // hook in the pre-commit config template. depName matches the `- repo:`/`rev:` entry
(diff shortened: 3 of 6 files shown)


pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
```diff
@@ -103,24 +103,59 @@ repos:
...
+        language: golang
+        additional_dependencies: ['github.com/giantswarm/schemalint/v2@v2.6.1']
+        files: ^helm/{{.}}/(values\.yaml|\.schema\.yaml|values\.schema\.json)$
+        pass_filenames: false
+        require_serial: true
+        entry: sh -c
         args:
-          - --config
-          - helm/{{.}}/.schema.yaml
+          - 'helm schema --config helm/{{.}}/.schema.yaml && python3 -c ''import json,sys; h=lambda o: ({**{k:v for k,v in o.items() if k!="additionalProperties"},"unevaluatedProperties":False}) if ("$ref" in o and o.get("additionalProperties") is False) else o; p=sys.argv[1]; d=json.load(open(p),object_hook=h); f=open(p,"w"); json.dump(d,f,indent=4); f.write(chr(10))'' helm/{{.}}/values.schema.json && schemalint normalize helm/{{.}}/values.schema.json -o helm/{{.}}/values.schema.json --force'
```
marians: **`unevaluatedProperties` is a hard error in schemalint's `cluster-app` ruleset — which devctl itself generates gates for.**

`schemalint`'s `AvoidUnevaluated` rule is part of the `cluster-app` ruleset with `SeverityError` (`pkg/lint/rules/avoid_unevaluated.go`: *"Property must not use unevaluatedProperties"*, `GetSeverity() -> SeverityError`).

devctl generates two gates that run exactly that ruleset:
- `pkg/gen/input/makefile/internal/file/Makefile.gen.cluster_app.mk.template:17` — `schemalint verify $(VALUES_SCHEMA) --rule-set=cluster-app` (`make validate-schema`)
- `cluster_app_schema_validation.yaml.template` → `giantswarm/github-workflows/.github/workflows/json-schema-validation.yaml`, which calls `giantswarm/schemalint/actions/verify-helm-schema` with `rule-set: 'cluster-app'`

Reproduced with the real v2.6.1 binary on a schema fixed by this hook:

```
$ schemalint verify helm/x/values.schema.json                       # generated hook -> OK
[SUCCESS] Input is valid JSON Schema.  [SUCCESS] Input is normalized.

$ schemalint verify helm/x/values.schema.json --rule-set=cluster-app
Errors (8)
Property must not use unevaluatedProperties
[ERROR] Input is not valid according to rule set 'cluster-app'.   # exit 1
```

So any repo generated with both the `cluster-app` flavour and the `helmchart` pre-commit flavour gets a `values.schema.json` that this hook makes pass `pre-commit` and fail the `JSON schema validation` workflow / `make validate-schema` — the same "no state satisfies both checks" class of problem the PR is fixing, just moved one layer out. The generated `schemalint-verify` hook here is safe only because it passes no `--rule-set`.

Worth deciding explicitly: either exclude the `$ref` fix when `cluster-app` is in play, or get `AvoidUnevaluated` relaxed/scoped upstream in schemalint before this lands broadly.

renovate-custom.json5
```diff
@@ -63,6 +63,20 @@
...
+    // schemalint pinned in `additional_dependencies` of the local helm-schema pipeline
+    // hook in the pre-commit config template. depName matches the `- repo:`/`rev:` entry
+    // for the same tool above, so Renovate consolidates both into one update and the
+    // normalizing binary never drifts from the verifying one -- a version skew between
+    // them could reintroduce the rival-formatter conflict (giantswarm/giantswarm#37267).
+    {
+      customType: 'regex',
+      managerFilePatterns: ['/pre-commit-config\\.yaml\\.template$/'],
+      matchStrings: [
+        'additional_dependencies: \\[\'github\\.com/giantswarm/schemalint/v2@(?<currentValue>[^\']+)\'\\]',
```
marians: **A major schemalint bump will produce an uninstallable `additional_dependencies` pin.**

The capture group only covers the version, not the `/v2` module-path suffix. With `datasource: github-tags` on `giantswarm/schemalint`, a `v3.0.0` tag makes Renovate rewrite the line to:

```yaml
additional_dependencies: ['github.com/giantswarm/schemalint/v2@v3.0.0']
```

which `go install` rejects (`module github.com/giantswarm/schemalint/v2@v3.0.0: invalid version: … should be v2`). Because this manager deliberately shares `depName` with the `- repo:`/`rev:` manager, the two are bumped in *one* commit — so the `rev:` goes to v3.0.0 and the pin goes broken at the same time, and the breakage only surfaces downstream: every generated chart repo fails at pre-commit env install, not in devctl CI.

The `rev:`-only pin on `main` did not have this failure mode; it's new with the module-path pin. Consider making the major part of the path part of the match, e.g.

```json5
      matchStrings: [
        'additional_dependencies: \\[\'github\\.com/giantswarm/schemalint/(?<depName>v\\d+)@(?<currentValue>[^\']+)\'\\]',
      ],
```

(or `datasource: go` with `depName: github.com/giantswarm/schemalint/v2`, accepting that consolidation with the `rev:` is lost), plus a rule that a schemalint major needs a manual template edit.

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
```diff
@@ -103,24 +103,59 @@ repos:
...
-      - id: helm-schema
+      - id: helm-schema-{{.}}
+        name: "generate + fix + normalize helm/{{.}}/values.schema.json"
+        language: golang
+        additional_dependencies: ['github.com/giantswarm/schemalint/v2@v2.6.1']
+        files: ^helm/{{.}}/(values\.yaml|\.schema\.yaml|values\.schema\.json)$
+        pass_filenames: false
+        require_serial: true
+        entry: sh -c
         args:
```
marians, line 156: **`open()` without `encoding=` makes the fix step locale-dependent.**

`json.load(open(p), …)` decodes with `locale.getpreferredencoding(False)`. On a machine/container where `LC_ALL=C` or `LC_CTYPE=C` is set explicitly (PEP 538 locale coercion is skipped when `LC_ALL` is set), that is ASCII, so the step dies with a `UnicodeDecodeError` traceback the moment a chart's schema contains a non-ASCII character. That is not hypothetical: descriptions come from `values.yaml` helm-docs comments (`useHelmDocs: true`) and en/em dashes and typographic quotes show up there routinely. Note `helm schema` writes raw UTF-8 (Go's encoder does not escape non-ASCII), so the bytes reaching Python really can be non-ASCII.

Cheap fix — also closes the leaked read handle:

```suggestion
          - 'helm schema --config helm/{{.}}/.schema.yaml && python3 -c ''import json,sys; h=lambda o: ({**{k:v for k,v in o.items() if k!="additionalProperties"},"unevaluatedProperties":False}) if ("$ref" in o and o.get("additionalProperties") is False) else o; p=sys.argv[1]; f=open(p,encoding="utf-8"); d=json.load(f,object_hook=h); f.close(); f=open(p,"w",encoding="utf-8"); json.dump(d,f,indent=4); f.write(chr(10)); f.close()'' helm/{{.}}/values.schema.json && schemalint normalize helm/{{.}}/values.schema.json -o helm/{{.}}/values.schema.json --force'
```

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
```diff
@@ -103,24 +103,59 @@ repos:
...
+  # cannot emit it and does not drop the bogus `additionalProperties` when annotated. So
+  # for every object that has both `$ref` and `additionalProperties: false` we drop
+  # `additionalProperties` and set `unevaluatedProperties: false`.
+  # Upstream bug: https://github.com/losisin/helm-values-schema-json/issues/317
+  #
+  # This replaces both the external `losisin/helm-values-schema-json` hook and the
+  # separate `schemalint-normalize` hook. `schemalint` is installed by the hook itself
+  # (`additional_dependencies`, hence `language: golang`); `helm schema` needs the `schema`
+  # helm plugin, installed in the pre-commit CI workflow -- the same requirement the
+  # external hook had, since it just ran `helm schema`.
```
marians: Small correction to "the same requirement the external hook had, since it just ran `helm schema`": the external hook did not *just* run `helm schema` — its entry was `scripts/git-hook.sh`, which guards first:

```sh
if ! (helm plugin list | grep -q "^schema\s" > /dev/null 2>&1); then
    echo "Please install helm-values-schema-json plugin! https://github.com/losisin/helm-values-schema-json#install"
    exit 1
fi
helm schema "${@}"
```

The plugin requirement is indeed unchanged, but the actionable message is lost. A dev without the plugin now sees `Error: unknown command "schema" for "helm"` from a hook named "generate + fix + normalize …", with no hint about what to install. Since every chart repo will hit this exactly once (first clone), it's worth keeping a one-line guard in front of the chain, e.g. `helm plugin list | grep -q "^schema[[:space:]]" || { echo "install helm-values-schema-json: https://github.com/losisin/helm-values-schema-json#install"; exit 1; } && helm schema …`.

marians (commented):
Reviewed the diff for correctness. I verified the core mechanics hold up — built the real `schemalint` v2.6.1 binary and ran the pipeline: the YAML/shell/Python quoting is correct (`$ref` survives, `sys.argv[1]` is the schema path), `schemalint normalize` round-trips through `orderedmap` so `unevaluatedProperties` is preserved, the output ends with a trailing newline, and `normalize(fix(generate()))` is a byte-identical fixed point with `schemalint verify` exiting 0. `go install github.com/giantswarm/schemalint/v2@v2.6.1` works (main package at module root, binary named `schemalint`), local `language: golang` hooks with `additional_dependencies` are supported down to the declared `minimum_pre_commit_version: '2.17'`, both Renovate regexes match the template, and `go test ./pkg/gen/...` passes.

Four issues, one of which I think needs a decision before this lands broadly (the `cluster-app` ruleset conflict).

renovate-custom.json5
marians: **Renovate: a major schemalint bump produces an invalid Go module path.** This manager rewrites only the captured version, keeping the literal `/v2@` prefix. When schemalint releases `v3.0.0`, `github-tags` will offer it and Renovate will write `github.com/giantswarm/schemalint/v2@v3.0.0` — an invalid module path (`go install` fails: *version v3.0.0 invalid: should be v2, not v3*), breaking the hook's install for every downstream repo. The sibling `rev:` manager would happily accept `rev: v3.0.0`, so the consolidated update lands half-broken — defeating the "never drift" goal. Consider constraining to non-major updates (`matchUpdateTypes`/`allowedVersions`) or capturing the `/vN` segment too so it moves in lockstep.

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
marians: **Altitude: the whole generate → $ref-fix → normalize pipeline, including a ~400-char inline Python program, lives in one YAML/shell/Python string.** Three layers of quoting have to stay in agreement (the Python deliberately avoids single quotes and uses `chr(10)`), and the core `$ref` → `unevaluatedProperties` transform — the actual bug fix — has no unit test; `Test_HelmSchemaFixHook` only greps the rendered config for substrings, so a regression in the transform logic (e.g. matching the wrong nodes) would not be caught. A small committed, unit-tested script (or pushing the fix upstream into the generator/schemalint) would be more durable than an un-testable one-liner embedded in a template.

marians: **Dead formatting work:** the Python step does `json.dump(..., indent=4)` and `f.write(chr(10))`, but `schemalint normalize` immediately rewrites the file into its own canonical form (that is the whole point — normalize is the last writer). Only the key transformation matters here; the indentation and trailing-newline effort is discarded on every run. Harmless, but worth dropping to keep the one-liner minimal.

marians: **Minor DX regression:** this hook invokes `helm schema` directly, so if the helm `schema` plugin is missing it fails with a cryptic plugin error. The losisin hook it replaces ran `scripts/git-hook.sh`, which first checked for the plugin and printed an actionable message (*"Please install helm-values-schema-json plugin! https://…"*) before exiting. Consider a similar guard so local devs without the plugin get a clear message.

pkg/gen/input/precommit/precommit_test.go
```diff
@@ -98,6 +98,102 @@ func Test_New_WithHelmchartFlavor(t *testing.T) {
...
+	chartDir := filepath.Join(dir, "helm", "test-chart")
+	if err := os.MkdirAll(chartDir, 0755); err != nil {
+		t.Fatalf("setup failed: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(chartDir, "Chart.yaml"), []byte("name: test-chart\n"), 0644); err != nil {
+		t.Fatalf("setup failed: %v", err)
+	}
+
+	origDir, _ := os.Getwd()
+	if err := os.Chdir(dir); err != nil {
```
marians, line 125: **Test fragility:** `Test_HelmSchemaFixHook` mutates the process-wide working directory via `os.Chdir(dir)` (restored with `defer`). This is shared global state; if any test in this package runs with `t.Parallel()` or otherwise depends on cwd, this introduces ordering coupling and potential flakiness. Prefer passing the temp dir explicitly (as `renderConfig`/`New` already accept a working dir via config) instead of chdir'ing the whole process.

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
piontec: Fixed in 6e47597 — both handles now opened with `encoding="utf-8"` and closed explicitly. Reproduced your failure mode first, and the test now pins it: `Test_RefFixPython` runs the program under `LC_ALL=C LC_CTYPE=C PYTHONUTF8=0` on a schema whose description carries an en dash and typographic quotes. Before the fix that is a `UnicodeDecodeError: ascii codec cannot decode byte 0xe2`; after it, it passes and the description round-trips.

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
piontec: Fixed in 6e47597 — `indent=4` and the `chr(10)` write are gone, so the fix step only does the key transformation and `schemalint normalize` (the last writer) owns the formatting.

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
piontec: Fixed in 6e47597, and thanks for digging out the actual wrapper — the PR description was loose there. The chain now starts with a guard:

```sh
helm plugin list | grep -q "^schema[[:space:]]" || { echo "helm schema plugin missing, install it with: helm plugin install https://github.com/losisin/helm-values-schema-json"; exit 1; }; helm schema --config ... && ...
```

Verified both branches with a fake `helm` on `PATH`: plugin absent → the message above and exit 1, with the generator never invoked; plugin present → the full pipeline runs. `[[:space:]]` instead of `\s` so it does not depend on GNU grep. Template comment updated too, since it claimed the external hook "just ran `helm schema`".

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
piontec: Same as #discussion_r3656394598 — fixed in 6e47597 with a `helm plugin list` guard in front of the chain that prints the install command and exits 1.

piontec: Fixed the testability half in 6e47597: the program moved out of the template into a `refFixPython` const in `internal/file/precommit.go`, injected as `{{ $.RefFixPython }}`. The template still renders byte-identically (verified by building `devctl` from the branch, running `gen precommit --flavors helmchart` on a real chart dir and reading the `args[0]` back through a YAML parser), but the transform now has a real test rather than a substring grep.

`Test_RefFixPython` executes the exact generated program against a schema shaped like `helm schema` output and asserts:
- a node with `$ref` + `additionalProperties: false` → `unevaluatedProperties: false`, `additionalProperties` dropped, `$ref` and description intact;
- a node **without** `$ref` keeps its `additionalProperties: false` and gains no `unevaluatedProperties` — the "matching the wrong nodes" regression you named, which would silently widen every plain object in the schema;
- the same for the root object.

It skips if `python3` is absent. Still not a committed script: that would be a new generated file in every chart repo whose only caller is this hook. Upstreaming remains the real fix — losisin/helm-values-schema-json#317 is the tracking issue, and the template points at it.

pkg/gen/input/precommit/precommit_test.go
piontec: Not changed, and the premise is slightly off: `precommit.Config` has no working-dir field — `New()` hardcodes `workingDir := "."` (`precommit.go:25`) and `FindHelmCharts` resolves against it, which is why the chart discovery has to see the temp dir as cwd. The sibling `Test_New_WithHelmchartFlavor` on `main` already does exactly this chdir dance, so this is the package pattern rather than something new here, and no test in the package calls `t.Parallel()`.

Making it injectable means adding a `WorkingDir` to the production config that only tests ever set to anything but `"."`. Happy to do it as a separate cleanup that converts both tests, but I would rather not grow the generator API inside this fix.

renovate-custom.json5
piontec: Good catch, fixed in 6e47597 — this was a real new failure mode, and downstream-only, which is the worst place for it.

Took the `datasource: go` route rather than capturing `/vN`, because it makes the pin structurally unable to go invalid: a schemalint v3 lives at `github.com/giantswarm/schemalint/v3`, so the Go proxy for the `/v2` path can only ever offer v2.x — Renovate cannot write a version the path does not support.

```json5
matchStrings: [
  'additional_dependencies: \\[\'(?<depName>github\\.com/giantswarm/schemalint/v2)@(?<currentValue>[^\']+)\'\\]',
],
datasourceTemplate: 'go',
```

The consolidation you noted as the reason for the shared `depName` is preserved by a `packageRule` instead, so the two pins still land in one PR and cannot skew:

```json5
{
  matchDepNames: ['giantswarm/schemalint', 'github.com/giantswarm/schemalint/v2'],
  groupName: 'schemalint',
},
{
  matchDepNames: ['giantswarm/schemalint', 'github.com/giantswarm/schemalint/v2'],
  matchUpdateTypes: ['major'],
  dependencyDashboardApproval: true,
},
```

The second rule covers the other half of your point: a major needs a manual template edit (module path *and* `rev:`), so it now waits on the dependency dashboard instead of opening a PR that bumps `rev:` to v3 while leaving an uninstallable pin behind. Verified both regexes still match the template after the change, and that the file still parses.

piontec: Same issue as #discussion_r3656392178 — fixed in 6e47597: the pin is now tracked with `datasource: go` on the `/v2` module path (which can never offer a v3), consolidation with the `rev:` is preserved by a `groupName: schemalint` packageRule, and majors require dependency dashboard approval since they need a manual module-path edit.

pkg/gen/input/precommit/internal/file/pre-commit-config.yaml.template
piontec: Thanks for reproducing this against the real binary — the `AvoidUnevaluated` / `cluster-app` conflict is real. I don't think it's reachable from a repo that works today, though, so I've left the `$ref` fix unconditional. Reasoning, and I'd like your sign-off since you have more context on the cluster-app repos than devctl's templates give me:

**The two flavours run the same pipeline in opposite directions, so they were already mutually exclusive.**

| | `helmchart` (`gen precommit` flavor) | `cluster-app` (`gen` flavour) |
|---|---|---|
| `values.schema.json` | **generated** from `values.yaml` by `helm schema` | hand-maintained, the source of truth |
| `values.yaml` | hand-maintained, the source of truth | **generated** from the schema (`make generate-values` → `helm-values-gen`) |
| `helm/*/README.md` | **generated** from `values.yaml` (`helm-docs` hook) | **generated** from the schema (`make generate-docs` → `schemadocs`) |

A repo with both doesn't just get an `unevaluatedProperties` lint error — its hand-written `values.schema.json` gets overwritten from `values.yaml` on every commit, and two generators fight over `helm/*/README.md`. Both of those are true on `main` too: the external `losisin` hook ran the same `helm schema`, and the `helm-docs` hook is unchanged. The `.schema.yaml` with `noAdditionalProperties: true` that drives all of this is itself only emitted by the `helmchart` flavor, so a cluster-app repo has no input for `helm schema` to begin with.

So the ruleset error is downstream of a combination that already destroys the schema — this PR changes which failure you hit first, not whether the combination works. That's why I read it as out of scope here rather than "the same problem moved one layer out": there's no state of `main` where both checks pass either.

**On the two options you offered:**

- *Exclude the `$ref` fix when `cluster-app` is in play* — `gen precommit` can't see the `gen` flavour; its own flavors are only `bash`/`md`/`helmchart` (`cmd/gen/precommit/flag.go:21`). It would need a new flag threaded in to conditionally disable a fix, for a combination that cannot produce a working repo anyway. I'd rather not add that knob speculatively.
- *Relax/scope `AvoidUnevaluated` upstream* — worth doing on its own merits (any cluster app that wants `$ref` + closed objects hits the same wall, with or without devctl), but it isn't blocking here as long as the combination above is genuinely unused.

**Two things I can't verify from here, and where I'd defer to you:** whether any real repo has both `--flavors helmchart` on `gen precommit` and `Makefile.gen.cluster_app.mk`, and whether you'd like the mutual exclusion made explicit. If yes to either, the honest shape is a validation in `gen precommit` that refuses `helmchart` when `Makefile.gen.cluster_app.mk` is present, so the collision fails loudly at generation time instead of silently in a chart repo's CI. Happy to add that — as a follow-up, since it's a pre-existing conflict rather than something this PR introduces. Leaving this thread open for your call.

marians: I don't know whether there are any repos with that conflict.

I agree that a safeguard against this conflicting config would make sense. And since we think helm-schema is the way forward, we should soon explain this to the KaaS teams and probably let them migrate their cluster apps to use the helm-schema workflow, too.

marians (approved):

PR #2113, opened 2026-08-07
feat(gen/circleci): read the Node version from .nvmrc

marians:
## What

`gen circleci --language node` now reads the `cimg/node` tag from the repo's `.nvmrc` when it pins an exact version, falling back to devctl's baked-in default otherwise. Also settable explicitly via `--node-image-version`.

## Why

A Node repo repeats its Node version in several places — a Dockerfile `FROM`, an `actions/setup-node` step, the generated CI job — and only the last of those is devctl's. So Renovate proposes the rendered `image: cimg/node:<ver>` line in `.circleci/workflows.yml` as its own PR ([example](https://github.com/giantswarm/backstage/pull/2059)), which is wrong twice over:

1. It changes only the `image:` line, leaving the `node-build` cache keys at the old version. Those keys are salted with the node version *precisely* so a bump cannot restore stale-ABI native addons (better-sqlite3, isolated-vm) — merging it silently breaks that invariant.
2. align-files reverts it on the next run, since devctl still holds the old value.

Same class of problem as the architect orb, which repos already disable Renovate for.

Reading `.nvmrc` gives the repo **one** file to change: it drives local dev (nvm/fnm/asdf/volta), `setup-node` via `node-version-file`, and — through this probe — both the executor image and the cache-key salt, which now cannot disagree. Renovate's built-in `nvm` manager owns `.nvmrc` natively (datasource `node-version`, depName `node`), so the bump also groups with the repo's other node deps and is LTS-gated, neither of which a rendered `cimg/node` tag can express.

The alternative — a `gen.ci.node.imageVersion` key in `giantswarm/github` — was rejected: it puts the value in *another* repo, so Renovate in the consuming repo still couldn't own it. That moves a place rather than removing one.

## How

`detectNodeVersion()` in `cmd/gen/circleci/runner.go`, alongside the `os.Stat("Dockerfile")` and `detectPackageManager()` content-probes already there. Resolution order: flag → `.nvmrc` → `DefaultNodeImageVersion`.

Only an exact `major.minor.patch` is honoured. `.nvmrc` also accepts aliases (`lts/*`, `node`) and partial versions (`24`), none of which name a `cimg/node` tag — a repo using one keeps the default rather than rendering an image that cannot be pulled.

## Blast radius

None until a repo opts in. Every repo today lacks a `.nvmrc`, so generation is unchanged — pinned by the existing golden files, which are untouched and still pass.

`NodeImageVersion` is renamed to `DefaultNodeImageVersion` (it is now a default, not the value); it keeps its `// renovate:` annotation.

## Testing

- `Test_detectNodeVersion` — 13 cases covering exact versions, `v` prefix, comments, whitespace, missing/empty file, and each unusable `.nvmrc` form.
- `Test_NodeImageVersion` — asserts an override reaches the `image:` line **and** the cache-key salt together, and that the default does not leak through.
- Existing goldens unchanged and green (`go test ./...`).

### Checklist


pkg/gen/input/circleci/circleci.go
```diff
@@ -47,13 +47,22 @@ const (
...
+// setup-node step) and must not diverge from CI -- overrides this by committing
+// a .nvmrc, which the runner probes. That keeps the version in ONE place per
+// repo instead of trading a central default for N copies: Renovate's built-in
+// nvm manager owns .nvmrc natively (datasource node-version, depName "node"),
+// so it groups with the repo's other node deps and is LTS-gated, neither of
+// which a rendered cimg/node tag can express.
 //
 // renovate: datasource=docker depName=cimg/node
-const NodeImageVersion = "24.18.0"
+const DefaultNodeImageVersion = "24.18.0"
```
marians, line 75: **The `// renovate:` annotation on this constant is inert — nothing bumps it.**

The doc comment says "Baked in and Renovate-managed, for the same reason as the orb pins", but the orb pins are managed by an explicit custom manager in `renovate-custom.json5` whose regex is anchored to the *name*:

```
'// renovate: datasource=(?<datasource>\S+) depName=(?<depName>\S+)\n\s*const \w*OrbVersion = "(?<currentValue>[^"]+)"'
```

`\w*OrbVersion` cannot match `DefaultNodeImageVersion` (nor the old `NodeImageVersion`). I checked the rest of the chain: `renovate-presets:default.json5` (+ `config:recommended`) and `lang-go.json5` declare no custom manager with a `.go` `managerFilePatterns`, and the testdata manager only matches `architect: giantswarm/architect@…`. So no manager in this repo's effective config reads this line.

Concrete consequence, and it is the half of this PR's design that has no owner: every repo without a `.nvmrc` — i.e. all of them today, per the blast-radius section — is frozen on `24.18.0` forever. `cimg/node:24.19.0` has been on Docker Hub since 2026-08-03 and no PR exists. When 24.18.0 picks up a CVE, nothing opens a bump; the "a toolchain bump ships with a devctl release" mechanism the comment relies on does not run.

This predates the PR, but the PR renames the constant and re-asserts the claim, and the `.nvmrc` arm only makes sense next to a default that actually moves. Adding a manager here would fix both:

```json5
    {
      customType: 'regex',
      managerFilePatterns: ['/pkg/gen/input/circleci/circleci\\.go$/'],
      matchStrings: [
        '// renovate: datasource=(?<datasource>\\S+) depName=(?<depName>\\S+)\\n\\s*const \\w*ImageVersion = "(?<currentValue>[^"]+)"',
      ],
    },
```

(the golden fixtures in `testdata/*.workflows.yml` hard-code `cimg/node:24.18.0` too, so a bump also needs a fixture entry or a manual regen — same problem the architect-orb fixture manager already solves.)

pkg/gen/input/circleci/circleci_test.go
```diff
@@ -1350,6 +1350,44 @@ func Test_NodeResourceClass(t *testing.T) {
...
 
+// Test_NodeImageVersion verifies the Node job renders devctl's baked-in default
+// when a repo pins nothing, and honours a repo's own pin (detected from .nvmrc
+// by the runner) otherwise. The pin must reach the executor image AND the
+// node-build cache-key salt together: those two are the pair that must never
+// disagree, since the cached node_modules holds native addons whose ABI is tied
+// to the node version. Bumping only one is exactly the failure a per-file
+// dependency bot causes when it edits the rendered image line alone.
+func Test_NodeImageVersion(t *testing.T) {
+	const pinned = "24.19.0"
```
marians: **This assertion fails the moment the default is bumped to `24.19.0` — which is the next value it will take.**

`pinned` is `24.19.0`; `DefaultNodeImageVersion` is `24.18.0`, and `cimg/node:24.19.0` is the current tag on Docker Hub. As soon as anyone bumps the default to `24.19.0`, `pinned == DefaultNodeImageVersion`, so `contains(override, DefaultNodeImageVersion)` is trivially true and this fires with a message that describes the opposite of what happened ("default … leaked through despite override") — sending the next reader hunting for a fallback bug that does not exist.

Verified: setting `DefaultNodeImageVersion = "24.19.0"` on this branch makes `go test -run Test_NodeImageVersion ./pkg/gen/input/circleci/` fail exactly this way.

A version that can never be a real default keeps the leak check honest:

```suggestion
	const pinned = "99.0.0"
```

cmd/gen/circleci/runner.go
```diff
@@ -123,3 +133,40 @@ func detectPackageManager() string {
...
+// node-version-file, and -- through this probe -- the generated cimg/node tag
+// and the node-build cache-key salt, which are otherwise the copies most prone
+// to silent drift.
+//
+// Only an exact version is honoured. .nvmrc also accepts aliases ("lts/*",
+// "node") and partial versions ("24"), none of which name a cimg/node tag; a
+// repo using one of those keeps devctl's baked-in default rather than
+// rendering an image that does not exist. Comments and surrounding whitespace
+// are stripped, matching how nvm and Renovate's nvm manager read the file.
+func detectNodeVersion() string {
```
marians: **An unusable `.nvmrc` is indistinguishable from no `.nvmrc`, and nothing tells the repo owner.**

Both the "file absent" and "file present but not `major.minor.patch`" paths return `""`, which `New()` then silently replaces with `DefaultNodeImageVersion`. That is the correct *value* (better than rendering a tag that may not exist), but the silence is the problem: `lts/*` and a bare major are the two most common `.nvmrc` contents in the wild, so the likely first-contact experience is

1. repo commits `.nvmrc` containing `lts/*` (which `setup-node`'s `node-version-file` and nvm both honour, resolving to the newest 24.x),
2. runs align-files, sees `.circleci/workflows.yml` unchanged at `cimg/node:24.18.0`,
3. concludes the probe is broken, or worse does not look — and now CI runs a *different* Node than local dev and the Dockerfile, which is precisely the drift the doc comment above says this probe removes.

The runner has `r.logger` and `r.stderr` right there. Distinguishing "no `.nvmrc`" (expected, silent) from "`.nvmrc` present but unusable" (warn, naming the value and what form is required) would make the fallback discoverable at the moment it happens. Returning a `(string, error)` or a second bool from `detectNodeVersion` is enough to tell the two apart.

cmd/gen/circleci/runner.go
```diff
@@ -123,3 +133,40 @@ func detectPackageManager() string {
...
+		}
+		return version
+	}
+
+	return ""
+}
+
+// exactNodeVersionRE matches a fully-qualified Node version (major.minor.patch)
+// -- the only form that maps 1:1 onto a cimg/node tag.
+var exactNodeVersionRE = regexp.MustCompile(`^\d+\.\d+\.\d+$`)
```
marians, line 194: **`major.minor` is rejected, but `cimg/node:24.19` genuinely exists — so this regex is stricter than the stated rationale.**

The comment here, the `detectNodeVersion` doc above, the `--node-image-version` help text and the CHANGELOG all justify the strictness with "partial versions (`24`) name no `cimg/node` tag". That holds for a bare major (I checked: `cimg/node:24` 404s on Docker Hub) but not for `major.minor` — the tag list currently includes `24.19`, `24.18`, `24.18.1`, `24.17`, `24.16`, … alongside every patch tag. `major.minor` is a perfectly ordinary `.nvmrc` value.

So a repo with `.nvmrc` = `24.19` silently keeps `24.18.0` even though `cimg/node:24.19` would pull fine and would track patches automatically. Two defensible fixes; either is better than the current mismatch between claim and behaviour:

```suggestion
// exactNodeVersionRE matches the Node version forms that name a cimg/node tag:
// major.minor.patch and major.minor (cimg publishes both). A bare major does
// not exist as a tag, so it stays rejected.
var exactNodeVersionRE = regexp.MustCompile(`^\d+\.\d+(\.\d+)?$`)
```

Or keep `^\d+\.\d+\.\d+$` and correct the three prose spots to say "an alias or a version less specific than `major.minor.patch`" rather than asserting no such tag exists — otherwise the next person to touch this reads the comment as fact.

pkg/gen/input/circleci/circleci.go
```diff
@@ -47,13 +47,22 @@ const (
...
+// runs on when a repo does not pin its own. Baked in and Renovate-managed, for
+// the same reason as the orb pins: a toolchain bump ships with a devctl release
+// and reaches repos via align-files rather than drifting per repo.
+//
+// A repo that needs to own its Node version -- because the version is also
+// baked into artifacts devctl does not generate (a Dockerfile FROM, a
+// setup-node step) and must not diverge from CI -- overrides this by committing
+// a .nvmrc, which the runner probes. That keeps the version in ONE place per
+// repo instead of trading a central default for N copies: Renovate's built-in
+// nvm manager owns .nvmrc natively (datasource node-version, depName "node"),
```
marians, line 60: **Moving the version to the `node-version` datasource drops the guarantee that the rendered tag is pullable.**

Today the value comes from a constant Renovate is (meant to be) tracking with `datasource=docker depName=cimg/node`, so "a `cimg/node` tag with this name exists" is a precondition of the bump. After this change the value comes from `.nvmrc`, which Renovate's `nvm` manager bumps from the `node-version` datasource — Node releases, not CircleCI images. CircleCI publishes `cimg/node` behind upstream by hours to days (`cimg/node:24.19.0` landed 2026-08-03).

So: Renovate bumps `.nvmrc` to a freshly released Node the day it ships → align-files runs in that window → `.circleci/workflows.yml` renders `image: cimg/node:<not-yet-published>` → every job in that repo fails at container spin-up with a pull error, on `main`, until CircleCI catches up. Self-healing and single-repo, but the repo is fully red in the meantime and the cause is not obvious from a `manifest unknown`.

Worth naming in the doc comment at minimum. If you want the guarantee back, `renovate-custom.json5` in the *consuming* repo can carry a `packageRules` entry pinning `.nvmrc`'s `node` dep to the `cimg/node` docker datasource via `replacementName`/a custom manager — or the `major.minor` form (see the regex thread) sidesteps it entirely, since `cimg/node:24.19` exists before any given patch does.

CHANGELOG.md
```diff
@@ -511,7 +526,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- `gen workflows --release-workflow=auto-release` emits a push-based release flow: `.github/workflows/auto-release.yaml` + `cliff.toml` at repo root. On every push to `main` / a backport branch matching `release-[0-9]*` or `release-v[0-9]*` (so `release-2.x`, `release-v3.7.x` -- but not `release-notes` or other unrelated branches), the workflow runs `git-cliff --unreleased --bump --context` to compute the next semver from conventional commits since the latest reachable v*._._ tag and `gh release create --target $GITHUB_SHA` to produce the tag + GitHub Release atomically. `cliff.toml`'s `[remote.github].repo` is auto-detected from the consuming repo's `origin` git remote URL. Mirrors the canonical version already deployed in muster, mcp-kubernetes, klaus-operator, agentic-platform, agentic-platform-mcps, agentic-platform-ui, mcp-toolkit, telemetrydeck-go, mcp-prometheus, klausctl, klaus-oci, mcp-debug, mcp-oauth (13 repos).
```
marians: **Unrelated formatter pass corrupted this line's content.** `v*.*.*` became `v*._._` — the emphasis-marker rewrite (`*x*` → `_x_`) misfired on a glob that was never emphasis. The sentence now reads "the latest reachable `v*._._` tag", which is not a thing.

Same pass produced two more unintended edits outside this PR's scope:
- **line 272**: `  supported with --signing-config` lost its leading indentation, dedenting a continuation line out of its list item.
- **line 72**: a blank line inserted mid-list, splitting the `### Changed` bullets into two lists (loose-list rendering).

None of it belongs to the `.nvmrc` change. Suggest reverting the three so the diff is just the new entry:

```suggestion
- `gen workflows --release-workflow=auto-release` emits a push-based release flow: `.github/workflows/auto-release.yaml` + `cliff.toml` at repo root. On every push to `main` / a backport branch matching `release-[0-9]*` or `release-v[0-9]*` (so `release-2.x`, `release-v3.7.x` -- but not `release-notes` or other unrelated branches), the workflow runs `git-cliff --unreleased --bump --context` to compute the next semver from conventional commits since the latest reachable v*.*.* tag and `gh release create --target $GITHUB_SHA` to produce the tag + GitHub Release atomically. `cliff.toml`'s `[remote.github].repo` is auto-detected from the consuming repo's `origin` git remote URL. Mirrors the canonical version already deployed in muster, mcp-kubernetes, klaus-operator, agentic-platform, agentic-platform-mcps, agentic-platform-ui, mcp-toolkit, telemetrydeck-go, mcp-prometheus, klausctl, klaus-oci, mcp-debug, mcp-oauth (13 repos).
```

pkg/gen/input/circleci/circleci.go
marians: Confirmed and fixed in f539349.

Verified the gap: the manager regex is `const \w*OrbVersion`, and `git log -S` shows the constant has been touched by exactly one commit — #2052, which introduced it. No bump PR has ever existed.

Added two managers: `const \w*ImageVersion` for the constant, and one for the golden fixtures matching both `image: cimg/node:` and the `node-build-…-<ver>-` cache-key salt (the fixtures pin the salt too, so matching only the image line would leave the goldens red).

Proved it with a dry-run against this branch:

```
DEBUG: 6 flattened updates found: cimg/node × 6
DEBUG: Returning 1 branch(es)  →  renovate/cimg-node-24.x
"replaceString": "// renovate: datasource=docker depName=cimg/node\nconst DefaultNodeImageVersion = \"24.18.0\"",  "newValue": "24.19.0"
```

Six deps: the constant plus every occurrence across both fixtures, all in one branch. Previously zero.

pkg/gen/input/circleci/circleci_test.go
marians: Fixed in f539349 — `pinned` is now `99.0.0`, with a comment explaining why it has to be unreachable.

Reproduced your check first (setting the default to `24.19.0` made the old assertion fail exactly as described), then confirmed the new one passes under the same bump.

cmd/gen/circleci/runner.go
marians: Agreed — fixed in f539349.

`detectNodeVersion` now returns `(version, rejected string)`: `rejected` is empty when there is no `.nvmrc` (expected, stays silent) and carries the raw value when the file exists but is unusable. The runner reports it on `r.stderr`.

End-to-end against a scratch repo containing `.nvmrc` = `lts/*`:

```
warning: ignoring .nvmrc value "lts/*" -- the Node job needs an exact major.minor.patch (e.g. 24.19.0); falling back to 24.18.0
```

The table test now asserts `wantRejected` per case, so the silent path and the warned path cannot collapse into each other.

cmd/gen/circleci/runner.go
marians: You are right about the facts, and I verified them against Docker Hub:

```
cimg/node:24.19    -> 200
cimg/node:24.19.0  -> 200
cimg/node:24.18    -> 200
cimg/node:24       -> 404
```

So `major.minor` exists and only a bare major does not — the prose was wrong.

I took your second option (keep `^\d+\.\d+\.\d+$`, fix the prose) rather than the suggestion, because accepting `major.minor` would undercut the feature. The repo Dockerfile pins an exact patch, so a floating `.nvmrc` would drift from it — reintroducing the divergence this probe exists to remove — and it would coarsen the `node-build` cache-key salt, which is meant to be exact. It would also make the `Node.js` grouping in the consuming repo meaningless, since a floating `.nvmrc` has nothing to bump while the Dockerfile still does.

Fixed in f539349: the regex comment, the `detectNodeVersion` doc, the flag help and the CHANGELOG now all say the rejection is deliberate and note that `cimg/node:24.19` does exist. The rejection is no longer silent either (see the thread above), so a repo hitting it is told what form to use.

Happy to switch to `^\d+\.\d+(\.\d+)?$` if you would rather have the flexibility — say the word.

pkg/gen/input/circleci/circleci.go
marians: Good catch — this was in the plan risks and the consuming repo`s `renovate-custom.json5`, but not here where it belongs. Documented on the constant in f539349:

> That trade has one cost worth knowing. This constant is tracked against the docker datasource, so "a cimg/node tag with this name exists" is a precondition of any bump. A .nvmrc is tracked against node-version — Node releases, which CircleCI trails by hours to days. A repo whose .nvmrc is bumped inside that window renders an image that cannot be pulled yet, and every job fails at container spin-up with a manifest error until CircleCI catches up. It is self-healing and confined to that one repo, but the cause is not obvious from the error; a consuming repo that wants the guarantee back can hold .nvmrc on the docker datasource in its own renovate config.

Also in the CHANGELOG entry. Not taking the `major.minor` escape hatch, for the reasons in the regex thread.

CHANGELOG.md
marians: Entirely my fault — an editor formatter rewrote the file after my edit and I committed it without checking the diff.

Reverted in f539349 by restoring `CHANGELOG.md` from before my commit and re-inserting only the new entries via a non-formatting write. The diff is now `26 insertions(+), 0 deletions(-)`.

There were three more unrelated edits beyond the ones you listed — the same pass also rewrote `*tarballs*` and `*first*` to underscore emphasis on untouched entries. Verified all are gone:

```
v*.*.*  -> 3 occurrences (intact)
v*._._  -> 0
"a make *rule*"  -> restored
"  supported with --signing-config"  -> indentation restored
```

Same thing had happened to `renovate-custom.json5` when I added the managers (it rewrote every string in the file from single to double quotes); that diff is now +29/-0.

